PR #31, opened 2021-02-09
RFC: New testing framework

dnicolodi:
@blais This is a sketch of how I imagine the testing framework could look like. There is no documentation yet, but I think it is rather self explanatory. I'll write some soon and polish the rough edges of this proof of concept. Let me know what you think.

Thing I like are using the document content sha1 as file name for the expected output, which frees the user from having to think about filenames, and the possibility to specify the documents source directory and expected output files directory with environment variables. If they are not specified on the command line or as env vars, I picked default relative to the script file.

For testing I save the output of the importer to file and compare it with the expected output. It is very simple, but it works. I was thinking about parsing the expected file and doing some more semantic comparison, but I think it is overkill. If the need arises, it should not be difficult to add. What do you think?


examples/office/importers/acme/acme_pdf.py
```diff
@@ -45,18 +29,9 @@ def pdf_to_text(filename):
...
-                                stderr=subprocess.PIPE)
-    except (FileNotFoundError, PermissionError):
-        pipe = subprocess.Popen(['pdf2txt', filename],
-                                stdout=subprocess.PIPE,
-                                stderr=subprocess.PIPE)
-    stdout, stderr = pipe.communicate()
-    if stderr:
-        raise ValueError(stderr.decode())
-    return stdout.decode()
+    r = subprocess.run(['pdftotext', filename, '-'],
```
blais, line 32: Is it clean what happens if pdtotext isn't installed on the user system?

examples/office/importers/utrade/tests/expected/d60c05338c9bb13c44249d0c9adefdde4ea273a5.beancount
```diff
@@ -1,24 +1,25 @@
+;;; document filing path: Assets/US/UTrade/2015-02-25-utrade.UTrade20150225.csv
```
blais, line 1: What about file_date()?
That one's an important piece of data that is produced from the extractors, because it helps determine the output filename and date of the document / application.

blais, line 1: Maybe choose something a bit more "programmatic" than "document filing path".
Maybe: "file_name", matching the method name,  so it's more explicit?

Also, you could contemplate using  Custom directive. Your call, don't  care that much.

blais:
Nice work! You're an ace

Some quick comments (it was a busy day on the markets, preparing for the worse):
- I like the hash as filename
- Avoid environment variables at all costs, I promise you they're a plague we'd rather not have. If we can get creative we could do away from them

BTW today I imported something and had to deal with one of those "takes forever to extract" PDF files. It's so frustrating. On Linux, something's stupid, some of the files just take forever. This is why I'd like to have a filesystem cache to avoid repeated those, and even install some form of timeout to bail out after e.g. 10secs (anything that works should always take ~ 1-2 secs).

examples/office/importers/utrade/tests/expected/d60c05338c9bb13c44249d0c9adefdde4ea273a5.beancount
dnicolodi: I opted for combining file_account, file_date, and file_date into "document filing path" (I arbitrarily choose this wording, if you have a better one I can change it) as the ingest framework does, it seems natural.

dnicolodi: See above, this is really the relative path to which the document would be filed.

blais: I'd rather see them as separate fields. Also, think about the possibility of the file naming scheme changing in the future to not have the date on it.

examples/office/importers/acme/acme_pdf.py
dnicolodi: You get an exception, same if `pdf2txt` is not installed. I think it is very likely that `pdftotext` is present on a desktop installation, but I don't have data to support this. I don't think it is worth implementing conditional tests in this new mechanism.

dnicolodi:
>     * Avoid environment variables at all costs, I promise you they're a plague we'd rather not have. If we can get creative we could do away from them

That's funny to hear from someone that has a `$L` environment variable pointing to his Beancount ledger :-)

The env vars are only used if present and if the corresponding options are not specified on the command line. I thought that it could be nice if you keep all your test documents in the same place to do not have to repeat the path all the time. This is also the reason why, by default, documents that are not recognized by the importer are ignored, unless the `--strict` option is used, in this case they generate an error.

If you think they are not useful and are actually detrimental, I can remove support for them.

blais:
> > ```
> > * Avoid environment variables at all costs, I promise you they're a plague we'd rather not have. If we can get creative we could do away from them
> > ```
> 
> That's funny to hear from someone that has a `$L` environment variable pointing to his Beancount ledger :-)
> 

Haha :-)
Yes, but there is no code that depends on it... it only lives in my environment.


> The env vars are only used if present and if the corresponding options are not specified on the command line. I thought that it could be nice if you keep all your test documents in the same place to do not have to repeat the path all the time. This is also the reason why, by default, documents that are not recognized by the importer are ignored, unless the `--strict` option is used, in this case they generate an error.
> 
> If you think they are not useful and are actually detrimental, I can remove support for them.

I know I used to do that, many times in the past. I even have a few in Beancount - they're driving me crazy. They cause a lot of problems over time. Best to try to get creative and avoid them, as nobody reads the documentation. It's one of those explicit > implicit cases. I have a Makefile to invoke all the oft-invoked commands, so that cruft doesn't get built into the programs and lives all in one place, how things are connected together.

examples/office/importers/utrade/tests/expected/d60c05338c9bb13c44249d0c9adefdde4ea273a5.beancount
dnicolodi: Sure, I'll change this.

dnicolodi: Also, I thought about using a `custom` or `note` directive, but then I would need to come up with a date for the directive, and I could not think to any that made sense.

dnicolodi:
New iteration, now with verbosity levels and colored console output! I added some documentation in form of the command help text. It remains to be decided what to do when the importer identifies a document but there is no expected output file for it. Right now a FileNotFound exception is raised. Maybe something more user-friendly would be more appropriate.

blais:
Can't wait to try it out.
Go ahead and merge it whenever you feel satisfied about it.

dnicolodi:
I wanted to add some tests before merging. Because the tests would be mostly comparing output strings I though that doctests would fit nicely, but then discovered that pytest does not really integrate that well with doctests... I'm investigating the options.

blais:
SGTM

dnicolodi:
BTW, what is the reason for using `pytest`? It seems that all tests use the `unittest` assertion model and not the `pytest` one. There are some things that `pytest` gets right, but some others not so much. I am tempted to go back to plain `unittest`

blais:
I just pytest as the runner, it's even better if we don't depend on it at
all indeed, I'm with you there

dnicolodi:
I now merged this. @blais, please give it a try and report anything you do like.

blais:
Thanks Daniele!

PR #33, opened 2021-02-13
cache: Add disk cache decorator

dnicolodi:
This is still missing documentation, but I know that @blais is eagerly waiting for something like this, thus this early PR to review. There are example of usage in the added tests. By default the cache is keyed on the filename and expired on the file modification time. However, this can be overwritten with the `key` argument to the cache decorator. In the tests there is an example for using the input file SHA1 digest. All arguments to the decorated function are always part of the cache key.

Which directory to use for the cache on WIndows? On Unix-like systems `~/.cache/beangulp/` is a good choice, but I don't have idea what would be the equivalent on Windows. Also, what should be the behavior when the cache directory does not exist? The implementation proposed simply raises  `FileNotFoundError` exception. Another possibility, would to silently disable the cache.

### DIFF
--- a/beangulp/cache.py
+++ b/beangulp/cache.py
@@ -7,6 +7,13 @@ text, can be expensive.
 __copyright__ = "Copyright (C) 2016  Martin Blais"
 __license__ = "GNU GPLv2"
 
+import functools
+import os
+import pickle
+import sys
+
+from contextlib import suppress
+from hashlib import sha1
 from os import path
 
 import chardet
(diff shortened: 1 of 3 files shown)


dnicolodi:
I think that ignoring `FileNotFoundError`s when the cache is being written is the best way forward. I don't think writing possibly sensitive data to disk without the user being fully aware is a good idea. In this way the user needs to opt in to the caching creating the cache directory `~/.cache/beangulp/`.

PR #34, opened 2021-02-13
regression: Drop old importers testing code

dnicolodi:
I don't see reason to keep this code around. @blais, do you?


blais:
I haven't been able to test your new one yet; I think if it covers the cases indeed this should go.
Give me a bit of time to test it out with at least one importer and I'll merge this.

PR #35, opened 2021-02-15
testing: Removed assumptions on testing directory structure.

blais:
- Removed 'expected' directory, hardcoding it to the same directory (on purpose,
  less options, less surprises).

- Added types and doctrings to beangulp.testing.

- Fixed bugs with importers producing None for filename, date, etc.


beangulp/testing.py
```diff
@@ -1,73 +1,118 @@
...
                                 fromfile='imported.beancount')
     return list(diff)
 
 
-def run_importer(importer, f):
-    account = importer.file_account(f)
-    date = importer.file_date(f)
-    name = importer.file_name(f)
-    entries = extract.extract_from_file(f.name, importer, None, None)
+# TODO(blais): Make the FileMemo type public, so it can be used here.
```
dnicolodi, line 101: I would like to get rid of FileMemo, thus I would not bother.

beangulp/testing.py
```diff
@@ -76,90 +121,72 @@ def log(msg, level=0, **kwargs):
...
+        expected_filename = "{}.beancount".format(doc)
+        cached_file = cache.get_file(path.abspath(doc))
+
+        # Run the importer's identify() method.
+        if importer.identify(cached_file):
+            account, date, name, entries = run_importer(importer, cached_file)
+            log(f'  {cached_file}', 1)
+            if account and date and name:
+                # TODO(blais): Figure out how to produce a log entry when some
+                # of these are None.
```
dnicolodi, line 164: I thought that returning None for any of these is a bug in the importer. Isn't it?

dnicolodi (commented):
I haven't commented on the issue of the location of the test documents and expected output files. I would like to explain my rationale for the choice I did and think a bit more about use cases. I'll write a document that can serve as base for user documentation and we can reason a bit more about it. In the mean time I'll merge the other changes. Thank you for adding types!

beangulp/testing.py
```diff
@@ -76,90 +122,72 @@ def log(msg, level=0, **kwargs):
...
-        f = cache.get_file(os.path.abspath(doc))
-
-        if importer.identify(f):
-            account, date, name, entries = run_importer(importer, f)
-            log(f'  {os.path.relpath(path):}', 1)
-            log(f'  {account.replace(":", "/"):}/{date:%Y-%m-%d}-{name:}', 1)
+        log(f'* {doc}', nl=verbosity > 0)
+
+        # Compute the path to the expected output file.
+        expected_filename = "{}.beancount".format(doc)
```
dnicolodi, line 155: The reason for using the SHA1 is to avoid ambiguity between changes in the test document and changes to the importer code. The SHA1 can be added to the expected file header. Another reason for using an unintelligible file name is to convey the idea that the user is not expected to touch it.

beangulp/testing.py
```diff
@@ -76,90 +122,72 @@ def log(msg, level=0, **kwargs):
...
-
-        # compute the path to the expected output file
-        path = os.path.join(expected, sha1sum(doc) + '.beancount')
-        f = cache.get_file(os.path.abspath(doc))
-
-        if importer.identify(f):
-            account, date, name, entries = run_importer(importer, f)
-            log(f'  {os.path.relpath(path):}', 1)
-            log(f'  {account.replace(":", "/"):}/{date:%Y-%m-%d}-{name:}', 1)
+        log(f'* {doc}', nl=verbosity > 0)
```
dnicolodi, line 152: Referring to your email about removing unicode to support non-unicode systems: do we really need to care about those? I think Python 3 is "unicode by default". The only issues could be printing unicode on the Windows console, and Click solves that for us, and font coverage on old terminals. I think that this is not a problem for terminal emulators in the last 20 years. The only think I am not sure about is unicode coverage of the font used by the Linux console. Summarizing: I think using characters outside ascii allows for nicer interfaces and there is no real technical reason to stick to ascii only. However, I can live with this change.

beangulp/testing.py
```diff
@@ -76,90 +122,72 @@ def log(msg, level=0, **kwargs):
...
-    root = os.path.dirname(__main__.__file__)
-
-    if not documents:
-        documents = (os.path.join(root, 'tests','documents'), )
-
-    if not expected:
-        expected = os.path.join(root, 'tests', 'expected', importer.name())
-        if not os.path.exists(expected):
-            expected = os.path.join(root, 'tests', 'expected')
-
```
dnicolodi, line 120: Removing this removes support for my use case: I have a few importers and I want to make sure that each identifies all the files it should identify, but not others. For ensuring this, I put all my test documents for all the importers in the same folder:
```
beans/
  ledger.beancount
  import.py
  importers/
     foo.py
     bar.py
     baz.py
     tests/
       documents/
         # all documents here
       expected/
         foo/
           # expected output for importer foo
         ...
```
In this way if and only if an expected output file exists in `expected/foo/` the foo importer should identify it. If you flatten the `exported/` directory hierarchy, I don't see an equivalently easy way to check this.

dnicolodi:
BTW, I think you should use `git rebase` more often. The history of this branch is a mess! :-)

blais:
Document: I'll make a copy of the old one and we can update that.
Here:
http://furius.ca/beancount/doc/beangulp
I shared with you, edit ACLs.

beangulp/testing.py
```diff
@@ -1,73 +1,118 @@
...
     try:
-        with open(filepath, 'r') as f:
-            lines_expected = f.readlines()
+        with open(filepath, 'r') as infile:
+            lines_expected = infile.readlines()
     except FileNotFoundError:
         lines_expected = []
 
+    # TODO(blais): Is there a way to ignore whitespace here?
+    # The current printer is producing extra junk at the end of lines.
```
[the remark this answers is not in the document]
blais: You are right! I didn't fix my tests before porting. It's my old whitespace that was still there.
No need to add whitespace normalization.
I would be a nice touch, but too much trouble (like if difflib just had an option I'd enable it, but more than that is not worth the effort).

beangulp/testing.py
blais: SGTM

beangulp/testing.py
blais: For the first point: how does it avoid ambiguity? I don't understand. Adding the SHA1 to the expected file header would seem sensible, but this means that every change to the importers - even cosmetic changes - would require regenerating all the expected files. Not great.

For the second point, having an unintelligible filename just makes it harder to find the expected output. Plus users of this know the purpose of these files, I don't think it indicates anything useful. It's real nice to have the expected file right next to the downloaded file, right in the same directory.

beangulp/testing.py
blais: I'm with you, but inasmuch as it's free in cost - like here, a cosmetic nicety - I see no reason to make the program depend on it.

beangulp/testing.py
dnicolodi: > For the first point: how does it avoid ambiguity? I don't understand. Adding the SHA1 to the expected file header would seem sensible, but this means that every change to the importers - even cosmetic changes - would require regenerating all the expected files. Not great.

The SHA1 of the test document, not the SHA1 of the importer implementation.

beangulp/testing.py
blais: I think file_date() and file_name() were always optional. 
See the docstrings here:
https://github.com/beancount/beangulp/blob/master/beangulp/importer.py#L84

I'll update the types in my PR so that account isn't optional.

Sometimes there's just no date... what does one do?
Also, if the filename doesn't stand to get renamed, it's okay to return None and keep the name by default.
Let's keep these optional, to ease the burden on users.

It's high time we add types to these :-)
(Will update PR)

beangulp/testing.py
blais: Hmm, okay. Maybe we should provide an optional override for the directory name. I do think the default should be as straightforward as possible though - remember the audience.

blais: Done. I've reinserted the option, but it assume the same basenames as the original files.
The default remains to find the files in the same directory.
Good enough?

beangulp/testing.py
blais: So it's not about hashing the importer source... I guess I don't understand what problem using the SHA1 of the input file solves?
We need to be able to easily find and associate that file with its corresponding input.

beangulp/testing.py
dnicolodi: I would prefer a scheme in which we look for the structured directories first and fall back to a flat directory structure if these are not there. Also, I don't think that having testing documents and expected output files in separate directory is such a cognitive load that it cannot be handled by someone expected to use this code.

beangulp/testing.py
dnicolodi: The idea is to protect against accidental changes or overwrites of the test documents (which are even more likely if the recommended test file location is not a dedicated directory).

examples/importers/runtests.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+DATA=../importers_tests
+python3 acme_pdf.py   test $DATA/acme_pdf
```
dnicolodi, line 3: This should be `python3 acme_pdf.py   test $DATA` ideally.

dnicolodi: Also, while we are at it, can we name the importers `acme.py` and `utrade.py`?

examples/README.md
```diff
@@ -0,0 +1,41 @@
...
+http://furius.ca/beancount/doc/beangulp for a fuller documentation and
+discussion of these files.
+
+There are five directories demonstrated here:
+
+* `ledger`: This is a directory, typically a repository under source control,
+  which contains your Beancount ledger file(s) (e.g., `ledger.beancount`) and
+  import configuration ('ledger.import'), which is a Python file.
+
+* `documents`: This is directory which contains your past imported files.
```
dnicolodi, line 14: "`documents`: a directory which contains..." or "documents`: a directory containing..."

blais: Done

examples/importers/runtests.sh
blais: Yep. I like it
I always have multiple formats in my real ones somehow, but this does make the examples simpler.
Will update now.

beangulp/testing.py
blais: Oh I see. These should always be under source control, so I'm not too worried. Perhaps we can rename the files before they get overwritten by default, and have an option to disable that to the new generate command.

examples/README.md
dnicolodi: Forgot to commit?

examples/importers/runtests.sh
blais: done

beangulp/testing.py
dnicolodi: I was more thinking of someone obliterating by mistake his test document, not the expected output file. However, if we stick to the separate command for generating the expected output file, it makes sense to refuse overwriting unless a `--force` option is provided. I didn't do that in my implementation because the CLI wound have been awkward.

examples/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+* `ledger`: This is a directory, typically a repository under source control,
+  containing your Beancount ledger file(s) (e.g., `ledger.beancount`) and import
+  configuration ('ledger.import'), which is a Python file.
+
+* `documents`: This is directory which contains your past imported files.
+  Beangulps' filing command is able to automatically date and rename the files
+  there after you've finished updating your Beancount ledger.
+
+* `importers`: This is directory, typically a repository under source control,
```
dnicolodi, line 20: This sentence is a bit strange in English.

examples/ledger/ledger.import
```diff
@@ -1,27 +1,33 @@
 #!/usr/bin/env python3
```
dnicolodi, line 1: Why rename this python script `ledger.import` doesn't `import.py` make more sense? Also, the directory in which the script is should be in the python path, thus the form `from importers import utrade` should work without having messing with the path, which IMHO is better style (at the price of a `__init__.py` in the `importers` directory.

blais: ledger.import, import.py... I don't have a strong opinion. I have something like five ledger.* files myself. Either way. I'll leave it here for now, happy to rename later.

About the path: no strong opinion either. The reason I flattened it is to try to make it as simple as possible. I want the example to keep things so obvious there are the least amount of questions. We can change it.

examples/README.md
blais: Fixed. Thank you.

PR #39, opened 2021-02-15
New importers interface

dnicolodi:
@blais This implements what we discussed. I implemented an adapter from the old interface to the new, so not all importers need to be ported to the new interface at once. We can remove it and the supporting code after the first release of beangulp. I am not very well versed in Python typing, but I think that we could make `beangulp.Importer` a `typing.Protocol` class, I am not sure.


beangulp/importer.py
```diff
@@ -128,3 +128,55 @@ def file_date(self, file) -> Optional[date]:
...
+        return f'{self.__class__.__module__}.{self.__class__.__name__}'
+
+    def identify(self, path: str) -> bool:
+        return False
+
+    def extract(self, path: str, existing: data.Entries = None) -> data.Entries:
+        raise NotImplementedError
+
+    # file() support interface
+
```
blais, line 146: No need to mention it

beangulp/importer.py
```diff
@@ -128,3 +128,55 @@ def file_date(self, file) -> Optional[date]:
           (If no date is returned, the file creation time is used. This is the
           default.)
         """
+
+
+class Importer:
```
blais, line 133: I love it, but I would put it at the top of the module, transfer the comments over to this one, make the comments on the previous version just say "See Importer.XXX". We want to emphasize that this is the current one.

beangulp/importer.py
```diff
@@ -128,3 +128,55 @@ def file_date(self, file) -> Optional[date]:
...
+
+    @property
+    def name(self) -> str:
+        return f'{self.__class__.__module__}.{self.__class__.__name__}'
+
+    def identify(self, path: str) -> bool:
```
blais, line 139: Thoughout the codef I actively avoid using the variable name "path" because it clashes with the common idiom of "from os import path". Prefer "filename"?

beangulp/importer.py
```diff
@@ -128,3 +128,55 @@ def file_date(self, file) -> Optional[date]:
...
+        return False
+
+    def extract(self, path: str, existing: data.Entries = None) -> data.Entries:
```
blais, line 142: So we're not changing the interface here yet? :-)

beangulp/importer.py
```diff
@@ -128,3 +128,55 @@ def file_date(self, file) -> Optional[date]:
...
+        raise NotImplementedError
+
+    def date(self, path: str) -> Optional[date]:
+        raise NotImplementedError
+
+    def filename(self, path: str) -> Optional[str]:
+        raise NotImplementedError
+
+
+class Adapter(Importer):
```
blais, line 157: Please add a liberal docstring describing what this is for.
(Might be obvious to you and I, but still.)

beangulp/__init__.py
```diff
@@ -17,6 +17,7 @@
 from beangulp import file
 from beangulp import identify
 from beangulp import importer
+from beangulp.importer import Importer
```
blais, line 20: I love it. We should make all public symbols on the top-level eventually, so that all importer code for the most part ever just uses "import beangulp"

beangulp/__init__.py
```diff
@@ -85,9 +86,17 @@ def _identify(ctx, src):
     return ctx.identify(src)
 
 
+def _importer(imp):
```
blais, line 89: Please add docstring.
(google style is to insert a docstring unless the function is painfully trivial.)

blais (approved):
I love it. Just a few comments

beangulp/importer.py
dnicolodi: Given the context I don't think it makes a big difference, but I will rename. I try to avoid confusion between "file name" intended as the base path, and the complete file path. Thus I will probably name this filepath.

beangulp/importer.py
dnicolodi: The first iteration of this patch removed the "existing" argument, but I later put it back: my importers use it. The use I have is to use the existing entries to train my ML transactions classifier. We could split this into a mandatoy `extract(filepath)` method and an optional `process(new, existing)` method (maybe with a better name). Or the post extraction processing could be moved to a method of the `Ingest` object that can be specialized in a subclass by the user. Or it could be moved to a completely new tool. However, for all these solutions, it wold not be possible to map the new interface to the old. It may be not that bad to provide compatibility only in a best effort way. Thought?

dnicolodi:
Thank you for the review. I forgot to mention that this is to be considered work in progress to reason in a more concrete way on the new interface.

PR #45, opened 2021-02-19
Return post-hook entries from extract()

bayesianmind:
This enables additional scriptability. For example, I would like to loop
over non-duplicates and merge them into my ledger automatically.


dnicolodi:
I am sympathetic to the use case, but I don't think this is the right interface. Why run the printer on the imported entries when the result will be discarded? 

My plan is to break the import functionality down in a few methods of the `Ingest` class. Currently I'm thinking about having `extract()`, `filter()`, `print()` methods, but I don't have all the details sorted out yet.

dnicolodi:
I didn't mean to close the PR. It happened because I force pushed the master branch to recover the history. See mail to the mailing list. This needs a rebase now. `git rebase --help` has instructions on how to deal with this.

PR #51, opened 2021-02-27
Updated cache code to use the updated mimetypes sniffer.

blais:


dnicolodi:
The tests are not happy. At a quick look I don't fully understand why. However, the old cache code will be used only for backward compatibility once we switch to the new importers interface. I don't know if it changing the old cache code is a good idea (it could make incrementally porting to beangulp more difficult) or worth the effort.

blais:
I needed the change in order to port all my importers otherwise I got a
warning.

I'm also experiencing a crazy large number of errors in using the file
command, I'll have to look into them, haven't had time yet.

dnicolodi:
> I needed the change in order to port all my importers otherwise I got a warning.

Maybe enabling the warning in the test runner was premature. However, you should be able to turn them off using `python -Wignore::DeprecationWarning` or with the `PYTHONWARNINGS` environment variable. If you see the warnings in a "normal" run of the importers, there is something funny about your setup.

> I'm also experiencing a crazy large number of errors in using the file command, I'll have to look into them, haven't had time yet.

Is this after or before this change? In the first case, I don't think we introduced any change that could modify the behavior of the importers. The second case, would be a good reason not to merge this patch :-)

blais:
My latest commit fixes the issue.

blais:
My test fixing here was really summary. I really needed this to work, lots of consolidation this weekend, it's nearly 3am and I'm still at it

dnicolodi:
I still think it would be easier to leave the `FileMemo` class alone and suppress the warning if it annoys you. The use of `FileMemo` will go away in the new importers interface definition. I don't think it is very productive to have to tuch up your importers and the tests twice: one for the mimetypes change and in a more radical way to port to the new interface.

blais:
Warning doesn't annoy me, I have 10 accounts to update this weekend and it was broken, I needed an immediate fix, not a whole lot of time to think.
Not enough cycles

dnicolodi:
Unless I'm missing something, the breakage comes from the first commit in this PR, which was supposed to eliminate the warning, so I don't understand what's your point. I'm happy to help fix what is broken, but I haven't understood what is broken.

blais:
I have tons of breakage, I need to fix quick locally, tons other things to take care of
I'll just wait until the cache is done and revert this branch
no time to describe, there's too much
ignore it

PR #54, opened 2021-03-01
New Importer interface

dnicolodi:
This is a proposal for a new Import interface.

Compared to the existing interface it
- drops the use of the memory cache,
- changes `name` from a function to an attribute (implemented as a property, in the base class),
- renames some of the methods,
- adds the `sortkey` static method and the `sort()` and `filter()` methods.

The `sortkey()` static method is used by the base class implementation of the `sort()` method in the obvious way. The `sort()` method allows to customize sorting of the entries extracted by the importer, defaulting to the current sorting. The `filter()` method allows to post-process the extracted entries, for example to plug in auto categorization of transactions.

I am debated whether an explicit `deduplicate()` method, maybe using  `cmp()` method under the hood, should be added.

An alternative approach could be to remove deduplication and filrering from the ingest tool and move them to a separate tool.

Comments?

### DIFF
--- a/.github/workflows/install.yaml
+++ b/.github/workflows/install.yaml
@@ -32,3 +32,9 @@ jobs:
           python importers/acme.py test importers_tests/acme --verbose
           python importers/csvbank.py test importers_tests/csvbank --verbose
           python importers/utrade.py test importers_tests/utrade --verbose
+      - name: Run example ledger.import
+        run: |
+          cd examples
+          python ledger/ledger.import identify Downloads -v
+          python ledger/ledger.import extract Downloads
+          python ledger/ledger.import archive Downloads -n -o documents
--- a/.pylintrc
+++ b/.pylintrc
@@ -24,6 +24,7 @@ disable=
   too-many-arguments,
   too-many-branches,
   too-many-locals,
+  too-many-return-statements,
   too-many-statements,
   undefined-loop-variable,
   unsubscriptable-object,
--- a/beangulp/__init__.py
+++ b/beangulp/__init__.py
@@ -12,6 +12,7 @@ __license__ = "GNU GPLv2"
 
 import os
 import sys
+import warnings
 import click
 
 from beancount import loader
@@ -89,9 +90,7 @@ def _extract(ctx, src, output, existing, reverse, failfast, quiet):
         if failfast and errors:
             break
 
-    # Invoke hooks.
-    hooks = [extract.find_duplicate_entries] if ctx.hooks is None else ctx.hooks
-    for func in hooks:
+    for func in ctx.hooks:
         extracted = func(extracted, existing_entries)
 
     # Reverse sort order, if requested.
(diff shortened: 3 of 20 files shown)


blais:
I love it.
Merge it.

We do need a "stable" branch at this point though, or a tag, I love the velocity but I can't keep up.

What do we do about caching though?

dnicolodi:
There are measures in place to make the old `importerProtocol` keep working thus updating to the new interface can be done one importer at time. The base class should provide default implementations for the methods that are being added, thus the porting should be quite straightforward.

The idea is that caching will become opt-in from the importers code, with the `@cache` decorator applied to the relevant functions (or with the `@memoize` decorator for in-memory cache that I plan to add soon).

dnicolodi:
I implemented an interface very similar to what is described above. However, I am not completely convinced that adding one more method to keep the consistency between the main importer methods that take the document path as only argument is the right choice. @blais , WDYT?

blais:
> I implemented an interface very similar to what is described above. However, I am not completely convinced that adding one more method to keep the consistency between the main importer methods that take the document path as only argument is the right choice. @blais , WDYT?
> 
> The new interface is defined here
> 
> https://github.com/beancount/beangulp/blob/e20b5396ef9c6d93c450934bb61a20c940a5555d/beangulp/importer.py#L17
> 
> I still want to move deduplication to be an importer method, but this requires a deeper surgery and intersect with #64

I don't know. By not providing the entries to extract(), you're forcing extract() to create all the entries and then separately filter. I think it's more powerful to just provide the entries to extract(), so it can avoid even creating a bunch of them in the first place. 

Moreover, for some of my importers, the prior state of the inventory is needed, e.g. to compute the sign for options expirations. By forcing the extraction and brewing to be separate that makes it difficult to do that.

dnicolodi:
Incorporating all comments and ideas, the new interface now looks like this: https://github.com/beancount/beangulp/blob/8284d18e277d6434d9322039dc0142cc9e882ac5/beangulp/importer.py#L26

I am fairly happy about how this looks like and I would merge it. Please shout if there is something you don't like.

blais:
I would change the deduplicate() implementation to call out to a regular function, single function call, so that nobody has the temptation to insert other methods called from it.

I try to avoid concrete inheritance normally, and ideally this importer is just abstract methods or trivial implementations.

dnicolodi:
It can be done, actually I had implemented it in this way at the beginning, but then the need to pass the `cmp()` function around made me opt for an implementation directly in the class. The implementation is trivial enough to live in the ABC, IMHO.

I am not sure I understand the concern. What do you mean with "nobody has the temptation to insert other methods called from it"? Insert other methods where? In the `Importer` ABC or in importers implementations?

blais:
These concrete implementations grow and the next thing you know a few
months down the road you end up with a big fat concrete class with a bunch
of methods that need to be called in a specific order and what not. That's
the concern. That's a big part of the reason why OO doesn't work, it
results in essentially external conditions on ordering, mutation, and a
bunch of side effects. I'd much rather keep the importer a purely abstract
class. That has much greater survivability characteristics.

dnicolodi:
Sure, any programming paradigm requires some discipline to be effective. On the other hand, you can write horrible code in many different ways :-) Le me know if you like this implementation better.

dnicolodi:
@blais unless you have any other reserve on the interface or on the implementation, I am going to merge this and move on

dnicolodi:
Thank you Martin!

Issue #52, opened 2021-03-01
Is the support for multiple importers matching the same document useful?


dnicolodi:
Currently more than one importers can positively identify the same document: all of them are run in sequence to extract entries, the first importer to match and that return a non-None account for the document is used to file the document.

Is this really desired?

I think this was coded this way to have different importers perform different operations on the same files (ie one importer does the extraction and another provides information for document filing). However, I think it takes a quite contrived example to find a case where this saves a significant amount of complexity in the importers. It is also confusing and error prone for the users: an error when more than one importer identifies the same document would be much better.

blais:
You are correct, that's the reason it was devised that way. e.g., some CSV files contain multiple sections and you may want to have separate importers for each section. In practice, it is almost never the same that you really, really need that, so I'm happy to simplify and bring this back to a 1:1 mapping.

You have my blessing!

dnicolodi:
Implemented in #58.

umangv:
Just wanted to add a data point that PNC Bank lumps checking and savings accounts into the same OFX file. There seems to be no way to download separate OFX files.

I have made a change locally to allow doing this via a command-line flag (`--allow-multiaccount-files`) on `extract` and `identify` . Would you be opposed to me bringing this behavior back but guarded by a flag so that it isn't the default behavior? If not, I'll get my change covered by unit tests and send over a PR.

umangv:
Oh, I guess this isn't as big an issue as I thought: an importer is not restricted to generating transactions for a single account. So what I would need to do to make this work is to have a single importer handle multiple accounts in the same bank and then export all transactions from that file.

I guess that's a convoluted way of saying that the change in #58 makes sense to me: it's worth protecting against the possibility that two importers have overly broad match conditions and someone ends up with duplicate (possibly conflicting) transactions as part of an import. The case I was worried about (an institution that bundles transactions from multiple accounts into a single file) can be handled by writing a more complicated importer.

PR #58, opened 2021-03-07
Restructure the command line interfce implementation

### LINKED ISSUE #3: [PATCH] bean-extract does not report nonzero exit code if an importer fails
**[Original report](https://bitbucket.org/blais/beancount/issue/206) by Mikhail Gusarov (GitHub: [misha-ridge](https://github.com/misha-ridge)).**

----------------------------------------

If any of importers fail, bean-extract still exits with 0 exit code, which is not very convenient for scripting.

Attached patch (my hg-fu is too weak to make a pull request) fixes this issue.

dnicolodi:
This implements #52, and fixes #3.


dnicolodi:
This is to be considered work in progress. Working on implementing the same treatment for the `file` and `extract` commands I tweaked the implementation up a bit. The general structure remains the same though.

dnicolodi:
This has grown into a major reorganization of the code. I'm quite happy of the result. The implemented functionality should be the same but I find the code much easier now. There are many commits but the important ones are 3d47e97, 
b2bc8cf, 5816867.

blais:
LGTM except for one thing: I'd really much rather we keep the tests right next to their modules and not to a separate directory. All the rest of the Beancount code is that way and it breaks the convention.

dnicolodi:
I moved the test to the separate directory for two reasons: compatibility with `unittest` requires the tests module names to be prefixed with `test_` instead than to be suffixed with `_test` and because of the doctests with I don't know how to make fit in the naming scheme you propose.

I started looking into the possibility of running the tests with `unittest` because it turns out that the `pytest` runner is hiding several resource leaks. The tests run with the unittest test runner reported the leaks and failed. At a very superficial look, the leaks were only in the test code, but I am not sure yet, and because of the recent fixes you committed to fix some in the CSV parser I think it would have been nice to catch them early.

blais:
These are good reasons but at the same time, it'll be annoying to have different ways to do things. It's nicer to do it all the same across all projects IMO.
I'm rather used to pytest as a default test runner, it's pretty slick, but I'm a dinosaur.
Is it out of fashion?

dnicolodi:
I don't think pytest is out of fashion but there are more things I find annoying about it that interesting features. I think that in Python 3 unittest gained the few features that could have made using pytest interesting for me. All projects I work on have tests in a separate directory. pytest discovers tests named in the unittest format just fine. What about moving the test alongside the code but named with a `test_` prefix instead than a `_test.py` suffix?

dnicolodi:
I moved the tests next to the modules they are testing, except for the doctests, these I kept in the `bangulp/tests/` directory. I renamed the `file` command to `archive` as discussed.

beangulp/extract.py
```diff
@@ -1,184 +1,114 @@
...
+                meta = entry.meta.copy()
+                meta[DUPLICATE] = True
+                entry = entry._replace(meta=meta)
+            marked.append(entry)
+        ret.append((filepath, marked))
+    return ret
+
+
+def print_extracted_entries(extracted, output):
+    """Print extracgted entries.
```
blais, line 89: typo

blais:
> I don't think pytest is out of fashion but there are more things I find annoying about it that interesting features. I think that in Python 3 unittest gained the few features that could have made using pytest interesting for me. All projects I work on have tests in a separate directory. pytest discovers tests named in the unittest format just fine. What about moving the test alongside the code but named with a `test_` prefix instead than a `_test.py` suffix?

Thank you Daniele!

If we use a suffix when you list the files in alphabetical order the implementation and the test sit right next to each other and it makes it obvious what doesn't have a test. It's not that super important, just another convention. The rest of the Beancount universe is named that way... I don't care all that much, though I do like things to be more like each other than different from each other unless there's some reason for it.

If you really care a lot for it, name it the way you like.

dnicolodi:
I proposed the different naming because `test_*.py` is what `unittest` expects and what most projects use. I was not aware that there is a command line switch to modify the file name glob pattern used by `unittest`. It may have been added relatively recently. With the tests next to the sources I prefer the current naming.

To run the tests with unittest, `python -m unittest discover -p '*_test.py'` is not much longer than `python -m unittest` which would suffice with the conventional unittest test naming.

PR #64, opened 2021-03-16
Fix and extend deduplication

dnicolodi:
Two small commits on top of #58. @blais I think I found why deduplication does not work reliably for you. See commit 45bcdf0. Commit d81fa2b extends the deduplication mechanism to work for entries within the same extract run as proposed in #9. Despite my initial schepticism, I think the algorithm I found makes a lot of sense.

### DIFF
--- a/beangulp/__init__.py
+++ b/beangulp/__init__.py
@@ -84,18 +84,25 @@ def _extract(ctx, src, output, existing, reverse, failfast, quiet):
 
             # Extract entries.
             entries = extract.extract_from_file(importer, filename, existing_entries)
-            extracted.append((filename, entries))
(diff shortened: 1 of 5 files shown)


nicdumz:
https://github.com/nicdumz/beangulp/tree/dedup-fix is a rebase, and after testing I think it fixes most of my issues in #93

dnicolodi:
> I think it fixes most of my issues in #93

Which issues does it not fix?

nicdumz:
No other problems. I use this patch / branch locally for all my beancount / beangulp needs now.

redstreet:
Hi, is there a plan to merge these two commits into master? I'm running into the same issue.

PR #67, opened 2021-03-27
mimetypes: Add text/beancount to the recognized MIME types.

dnicolodi:
Associate .beancount and .beans extensions to text/beancount MIME type.


blais:
Wow do we have our own MIME type yet? ;-)

dnicolodi:
> Wow do we have our own MIME type yet? ;-)

While I think that applying for a one with the IANA https://www.iana.org/assignments/media-types/media-types.xhtml may be overkill, we could have one added to shared-mime-info https://gitlab.freedesktop.org/xdg/shared-mime-info/

PR #72, opened 2021-04-07
Add support for currency fields in CSV files

kubauk:
I would like to introduce functionality that will parse and understand currency fields in CSV files. Many of us have transactions in multiple currencies and there are more and more financial institutions that provide multi-currency accounts.

The most contentious change is in the last comment where I modify a condition from __balance is not None__ to __not balance__. This change means where before empty balance fields would produce a zero balance assertion, they will now produce none. 

This test seems more correct to me as it also test for empty strings as well as None. There are many other tests for other variables and fields that are still is not None which I think may have meant to be not <variable> but I have not addressed those in this PR.

### DIFF
--- a/beangulp/importers/csv.py
+++ b/beangulp/importers/csv.py
@@ -73,6 +73,10 @@ class Col(enum.Enum):
     # account.
     CATEGORY = '[CATEGORY]'
 
+    # A column that indicates the amount currency for the current row which may
+    # be different to the base currency.
+    CURRENCY = '[CURRENCY]'
+
 
 def get_amounts(iconfig, row, allow_zero_amounts, parse_amount):
     """Get the amount columns of a row.
@@ -275,6 +279,11 @@ class Importer(identifier.IdentifyMixin, filing.FilingMixin):
 
                 balance = get(row, Col.BALANCE)
 
+                if Col.CURRENCY not in iconfig or not get(row, Col.CURRENCY):
+                    currency = self.currency
+                else:
+                    currency = get(row, Col.CURRENCY)
+
                 # Create a transaction
                 meta = data.new_metadata(file.name, index)
                 if txn_date is not None:
@@ -283,7 +292,7 @@ class Importer(identifier.IdentifyMixin, filing.FilingMixin):
                 if txn_time is not None:
                     meta['time'] = str(dateutil.parser.parse(txn_time).time())
                 if balance is not None:
-                    meta['balance'] = self.parse_amount(balance)
+                    meta['balance'] = Amount(self.parse_amount(balance), currency)
                 if last4:
                     last4_friendly = self.last4_map.get(last4.strip())
                     meta['card'] = last4_friendly if last4_friendly else last4
@@ -304,7 +313,7 @@ class Importer(identifier.IdentifyMixin, filing.FilingMixin):
                         continue
                     if self.invert_sign:
                         amount = -amount
-                    units = Amount(amount, self.currency)
+                    units = Amount(amount, currency)
                     txn.postings.append(
                         data.Posting(account, units, None, None, None, None))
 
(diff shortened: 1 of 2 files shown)


dnicolodi:
The change regarding the balance directives seems wrong to me. The `is not None` check is used to distinguish between a value not set and a numeric value that may be zero thus false in boolean context. I haven't looked carefully at the code, but at that stage, `balance` should be a `Decimal` object, thus I don't understand the comment about the empty string.

Can you also please squash the commits into a single commit?

beangulp/importers/csv.py
```diff
@@ -274,6 +277,9 @@ def get(row, ftype):
                 last4 = get(row, Col.LAST4)
 
                 balance = get(row, Col.BALANCE)
+                currency = self.currency \
+                    if Col.CURRENCY not in iconfig or not get(row, Col.CURRENCY) \
+                    else get(row, Col.CURRENCY)
```
dnicolodi, line 282: It seems that this would be easier to unwrap into a regular `if ... else` block.

beangulp/importers/csv.py
```diff
@@ -73,6 +73,9 @@ class Col(enum.Enum):
     # account.
     CATEGORY = '[CATEGORY]'
 
+    # A column that indicates the amount currency for the current row which may be different to the base currency.
```
dnicolodi, line 76: This line is to long. Can you please wrap the comment a 78 columns or there about?

beangulp/importers/csv.py
```diff
@@ -327,14 +333,23 @@ def get(row, ftype):
...
+            entry = None
+            if Col.CURRENCY in iconfig:
+                for e in entries[::-1]:
+                    balance = e.meta.get('balance', None)
+                    if balance and balance[1] == self.currency:
+                        entry = e
+                        break
+
+            if not entry:
+                entry = entries[-1]
```
dnicolodi, line 345: I have the impression that this does not do what intended if there are no transactions in the main currency in the CSV file. Also, wouldn't it be better to emit one `balance` directive for each currency in the file?

kubauk: Yes I was also thinking about producing a balance per currency however I stopped at one because in my specific case I did not need it to do so. If you think it's required I can put the functionality in.

kubauk:
The balance change in an interesting one. For some reason, I had decided that sometimes CSV could have a balance field but the contents could be empty. In this case the change I added is needed as both None and '' are indications of no balance being present.

I'm now thinking I should revert the change and not cover this scenario as it really should not happen. Why would an institution put the field into the CSV without then populating it with valid data?

kubauk:
I've ended up adding support for multiple balance assertions and I'm actually much happier with the code now. It's far more readable and just seems like the right thing to do in this case. 

I've also reverted the change around the balance test on the basis of my last comment. I just don't see the case it was catering for to be a valid one.

beangulp/importers/csv.py
```diff
@@ -283,7 +292,7 @@ def get(row, ftype):
                 if txn_time is not None:
                     meta['time'] = str(dateutil.parser.parse(txn_time).time())
                 if balance is not None:
-                    meta['balance'] = self.parse_amount(balance)
+                    meta['balance'] = (self.parse_amount(balance), currency)
```
dnicolodi, line 295: This would look more idiomatic if written
```python
meta['balance'] = Amount(self.parse_amount(balance), currency)
```
and the rest of the code adjusted to this change.

beangulp/importers/csv.py
```diff
@@ -325,16 +334,22 @@ def get(row, ftype):
         if not is_ascending:
             entries = list(reversed(entries))
 
+        balances = {}
         # Add a balance entry if possible
-        if Col.BALANCE in iconfig and entries:
-            entry = entries[-1]
+        for entry in entries[::-1]:
```
dnicolodi, line 339: I think the `if Col.BALANCE in iconfig` part should stay, to avoid iterating the entries if it is known that there are no balances.

kubauk: That make sense. I will change it back to include the test.

beangulp/importers/csv.py
kubauk: I missed this in the last push. I will add the changes now.

dnicolodi:
I now committed a slightly cleaned up version of this patch. Please test and let me know.

PR #79, opened 2021-05-09
Fix bug in deduping logic

doriath:


dnicolodi:
Thank you. This feature really needs some more tests.

PR #80, opened 2021-05-12
Add option to set has_header to fix one-line csv bug

ThomasdenH:
If a csv file is parsed without a header that contains only one entry, the current importer would interpret it as a header. This PR gives an additional setting to circumvent the automatic detection of a header.

### DIFF
--- a/beangulp/importers/csv.py
+++ b/beangulp/importers/csv.py
@@ -128,6 +128,7 @@ class Importer(identifier.IdentifyMixin, filing.FilingMixin):
                  narration_sep: str = '; ',
                  encoding: Optional[str] = None,
                  invert_sign: Optional[bool] = False,
+                 has_header: Optional[bool] = None,
                  **kwds):
         """Constructor.
 
@@ -149,6 +150,8 @@ class Importer(identifier.IdentifyMixin, filing.FilingMixin):
             narration fields of a source field.
           encoding: Encoding for the file, utf-8 if not specified or None.
           invert_sign: If true, invert the amount's sign unconditionally.
+          has_header: An optional boolean to indicate whether the csv file has a header.
+            If not set, the importer will try to detect it automatically.
           **kwds: Extra keyword arguments to provide to the base mixins.
         """
         assert isinstance(config, dict), "Invalid type: {}".format(config)
@@ -164,6 +167,7 @@ class Importer(identifier.IdentifyMixin, filing.FilingMixin):
         self.narration_sep = narration_sep
         self.encoding = encoding or 'utf-8'
         self.invert_sign = invert_sign
+        self.has_header = has_header
 
         self.categorizer = categorizer
 
@@ -192,6 +196,7 @@ class Importer(identifier.IdentifyMixin, filing.FilingMixin):
             file.head(encoding=self.encoding),
             self.csv_dialect,
             self.skip_lines,
+            self.has_header
         )
         if Col.DATE in iconfig:
             with open(file.name, encoding=self.encoding) as infile:
@@ -222,6 +227,7 @@ class Importer(identifier.IdentifyMixin, filing.FilingMixin):
             file.head(encoding=self.encoding),
             self.csv_dialect,
             self.skip_lines,
+            self.has_header
         )
 
         with open(file.name, encoding=self.encoding) as infile:
@@ -377,7 +383,7 @@ class Importer(identifier.IdentifyMixin, filing.FilingMixin):
         return get_amounts(iconfig, row, allow_zero_amounts, parse_amount)
 
 
-def normalize_config(config, head, dialect='excel', skip_lines: int = 0):
+def normalize_config(config, head, dialect='excel', skip_lines: int = 0, has_header: Optional[bool] = None):
     """Using the header line, convert the configuration field name lookups to int indexes.
 
     Args:
@@ -399,8 +405,8 @@ def normalize_config(config, head, dialect='excel', skip_lines: int = 0):
 
     head = io.StringIO(head, newline=None)
     lines = list(head)[skip_lines:]
-
-    has_header = csv.Sniffer().has_header('\n'.join(lines))
+    if has_header == None:
+        has_header = csv.Sniffer().has_header('\n'.join(lines))
     if has_header:
         header = next(csv.reader(lines, dialect=dialect))
         field_map = {name.strip(): index for index, name in enumerate(header)}
--- a/beangulp/importers/csv_test.py
+++ b/beangulp/importers/csv_test.py
@@ -40,6 +40,14 @@ class TestCSVFunctions(unittest.TestCase):
         iconfig, _ = csv.normalize_config({Col.DATE: 1}, head)
         self.assertEqual({Col.DATE: 1}, iconfig)
 
+    def test_normalize_config__has_header_false(self):
+        head = textwrap.dedent("""\
+          DEBIT,3/18/2016,"Payment to Chafe card ending in 1234 03/18",-2680.89,ACCT_XFER,3409.86,,
+        """)
+        iconfig, has_header = csv.normalize_config({Col.DATE: 1}, head,
+                                                   skip_lines=4, has_header=False)
+        self.assertFalse(has_header)
+
     def test_normalize_config__with_skip_and_header(self):
         head = textwrap.dedent("""\
           Lorem ipsum dolor sit amet, consectetur adipiscing elit.
@@ -177,6 +185,25 @@ class TestCSVImporter(cmptest.TestCase):
 
         """, entries)
 
+    @test_utils.docfile
+    def test_one_line_no_header(self, filename):
+        """\
+          11/7/2016,A,2
+        """
+        file = cache.get_file(filename)
+        importer = csv.Importer({Col.DATE: 0,
+                                 Col.NARRATION: 1,
+                                 Col.AMOUNT: 2},
+                                'Assets:Bank', 'EUR', [],
+                                dateutil_kwds={'dayfirst': True},
+                                has_header=False)
+        entries = importer.extract(file)
+        self.assertEqualEntries(r"""
+
+          2016-07-11 * "A"
+            Assets:Bank  2 EUR
+
+        """, entries)
 
     @test_utils.docfile
     def test_links(self, filename):


beangulp/importers/csv.py
```diff
@@ -149,6 +150,8 @@ def __init__(self, config, account, currency,
             narration fields of a source field.
           encoding: Encoding for the file, utf-8 if not specified or None.
           invert_sign: If true, invert the amount's sign unconditionally.
+          has_header: An optional boolean to indicate whether the csv file has a header.
+            If not set, the importer will try to detect it automatically.
```
dnicolodi, line 154: I don't particularly like the name of this argument. It is an argument passed to the importer constructor but the the think that has or not an header are the parsed files. I would much prefer a `names` that takes values `True` to force the first line of the file to be interpreted as column names, `False` to force it not to be interpreted as column names, `None` for auto detection as currently done, or a list of strings to do not interpret the first line of the file as column names but use the provided names instead.

beangulp/importers/csv.py
```diff
@@ -399,8 +405,8 @@ def normalize_config(config, head, dialect='excel', skip_lines: int = 0):
 
     head = io.StringIO(head, newline=None)
     lines = list(head)[skip_lines:]
-
-    has_header = csv.Sniffer().has_header('\n'.join(lines))
+    if has_header == None:
```
dnicolodi, line 408: Checking for `None` values should be done with `is None`.

beangulp/importers/csv_test.py
```diff
@@ -40,6 +40,14 @@ def test_normalize_config__with_header(self):
         iconfig, _ = csv.normalize_config({Col.DATE: 1}, head)
         self.assertEqual({Col.DATE: 1}, iconfig)
 
+    def test_normalize_config__has_header_false(self):
+        head = textwrap.dedent("""\
+          DEBIT,3/18/2016,"Payment to Chafe card ending in 1234 03/18",-2680.89,ACCT_XFER,3409.86,,
+        """)
+        iconfig, has_header = csv.normalize_config({Col.DATE: 1}, head,
+                                                   skip_lines=4, has_header=False)
```
dnicolodi, line 48: You are skipping 4 lines in a file that has only one. I don't think this is testing what you think ti is testing.

beangulp/importers/csv_test.py
```diff
@@ -177,6 +185,25 @@ def test_date_formats(self, filename):
...
+    def test_one_line_no_header(self, filename):
+        """\
+          11/7/2016,A,2
+        """
+        file = cache.get_file(filename)
+        importer = csv.Importer({Col.DATE: 0,
+                                 Col.NARRATION: 1,
+                                 Col.AMOUNT: 2},
+                                'Assets:Bank', 'EUR', [],
+                                dateutil_kwds={'dayfirst': True},
```
dnicolodi, line 198: Please keep the tests as focused as possible to the aspect they want to tests. In other words, there is no need to test the `dateutil_kwds` argument here.

beangulp/importers/csv_test.py
```diff
@@ -177,6 +185,25 @@ def test_date_formats(self, filename):
...
+                                has_header=False)
+        entries = importer.extract(file)
+        self.assertEqualEntries(r"""
+
+          2016-07-11 * "A"
+            Assets:Bank  2 EUR
+
+        """, entries)
```
dnicolodi, line 206: No need for all this empty lines.

dnicolodi (changes requested):
The ratio of options to number of lines of code is approaching one, but I guess an option on the lines of what you propose makes sense. However, I think a more general approach would be better, see the other comments.

beangulp/importers/csv.py
ThomasdenH: What would happen if the automatic detection detects a header but `use_names` is set to `False`? Would the first line still be read? If only column indexes are used, the variable would then only change whether the first line is skipped or not. So you could end up in the situation where you'd have to set `use_names=True` to skip the header even if you don't use names.

`skip_lines` could also be used for that purpose, but to me that's to skip a preceding section that is not arranged in columns.

Maybe a better solution is to never skip the header but require the user to account for it with `skip_lines` or if not set, set `skip_lines` in `normalize_config` to 1 or 0 depending on whether a header was detected.

dnicolodi: > What would happen if the automatic detection detects a header but `use_names` is set to `False`?

With `names` set to `False` there is no automatic detection.

> Maybe a better solution is to never skip the header but require the user to account for it with `skip_lines` or if not set, set `skip_lines` in `normalize_config` to 1 or 0 depending on whether a header was detected. That would fix the bug and wouldn't even require a new argument.

This would break backward compatibility and it is much clearer if`skip_lines` means skip not "maybe interpret them as column names".

ThomasdenH: > > Maybe a better solution is to never skip the header but require the user to account for it with `skip_lines` or if not set, set `skip_lines` in `normalize_config` to 1 or 0 depending on whether a header was detected. That would fix the bug and wouldn't even require a new argument.
> 
> This would break backward compatibility and it is much clearer if`skip_lines` means skip not "maybe interpret them as column names".

I was probably not clear. That's what I meant, skip exactly `skip_lines` regardless of their interpretation. Currently it means 'skip `skip_lines` and then skip an additional line if a header is detected', which may fail if the detection is incorrect. That can be solved by overriding the detection or not skipping an extra line conditionally. It does break backward compatibility.
 
> 
> > What would happen if the automatic detection detects a header but `use_names` is set to `False`?
> 
> With `names` set to `False` there is no automatic detection.
> 

The bug is that the first line is incorrectly seen as a header and skipped in the file. I think a single-line CSV will just be interpreted as a header. The name `names` may be confusing because when there is a header but no names are used in the configuration, the setting should still be `names=True`, not `names=False` as some might expect. I think you intend `names` to mean 'the first line of this file contains the column names' and not 'I want to use names to refer to columns in my configuration', but to me the name isn't clear. If backwards compatibility can't be broken maybe there's another name for the argument.

dnicolodi: > The bug is that the first line is incorrectly seen as a header and skipped in the file. I think a single-line CSV will just be interpreted as a header. The name `names` may be confusing because when there is a header but no names are used in the configuration, the setting should still be `names=True`, not `names=False` as some might expect.

If you want the first line to be ignored just set `skip_lines=1`. I don't see what is confusing about it. `names=True` means use the first line after `skip_lines` for columns names. `names=False` means don't use the first line after `skip_lines` as column names. Obviously in this case you cannot have column definitions based on column names. `names=None` means autodetect if the first line after `skip_lines` has column names or data, like it is done unconditionally now. `names=[...]` mean use exactly the provided names for column names and don't do anything special with the first line after `skip_lines`. This last possible value for the option would be nice to have and not hard to implement, but I would not insist on it.

The name of the option and its interpretation is based on the `numpy.genfromtxt` function.

Another possibility would be to have an `header` parameter. `header=None` means autodetect if the first line after `skip_lines` has column names or data, like it is done unconditionally now. `header=N` means use line N for columns names, ignore all lines up to that one. `header=False` means there is not line with column names. Obviously in this case you cannot have column definitions based on column names. This has the drawback that the interaction with `skip_lines` is ill defined: should `skip_lines=5, header=3` use line 8 for the column names or use line 3 for the header and start reading data at line 8?

This would be loosely modeled after the `pandas.read_csv` function. But Pandas is not the best source of inspiration for clear and concise APIs IMHO. I think the other alternative fits better.

dnicolodi:
@ThomasdenH I would appreciate if you have a look at #81 and see if that approach works for your CSV importers. It should allow to easily cover all use cases the current CSV importer covers and many more. For an example see the `newcsvbank.py` importer in the examples.

ThomasdenH:
Thanks! The new importer fits my needs well.

dnicolodi:
Thank you for having a look. I didn't mean that to be a replacement for fixing the existing importer and thus to this PR. However, I tend toward making `csvbase.Importer` the recommended solution, deprecate `csv.Importer` and remove it after the first release of beangulp. Do you have any comment on the extension interface of `csvbase.Importer`? I would appreciate some feedback before I merge it into beangulp.

ThomasdenH:
- I couldn't really figure out how to use the filing and identify mixins, but that is probably due to my limited experience with Python.
- I use one csv format where amounts use a comma as a decimal separator. It was very easy to add a column type to fix that, but it could also be a simple addition to the Amount type.
- The old formatter filtered out 0 value transactions (I think). This one doesn't, which is probably better, but I would still like some way to filter them out. For me it would be enough if `finalize` accepted `None` as a return type to indicate that a transaction should be removed.

dnicolodi:
> I couldn't really figure out how to use the filing and identify mixins, but that is probably due to my limited experience with Python.

This is by design. The mixins are considered an awkward API are are going to be phased out with the switch to the new `Importer` interface. I find it much easier to implement the `identify()` and the archiving related methods directly in the importer. If your experience is different I would like to hear about it.

> I use one csv format where amounts use a comma as a decimal separator. It was very easy to add a column type to fix that, but it could also be a simple addition to the Amount type.

I thought about it, but I failed in coming up with a general API that does not involve passing a parse function to the `Column` definition and if you do that you can as well define a new `Column` type:
```python
class Amount(csvbase.Amount):
    def parse(self, value):
        return super().parse(value.replace(',', '.')
```
Do you have in idea for a suitable API?

> The old formatter filtered out 0 value transactions (I think). This one doesn't, which is probably better, but I would still like some way to filter them out.

I think the old CSV importer did some filtering, indeed, but why would be transactions with zero amount be in the bank statements (or other source) if they do not contain any interesting information? I thus thought that the filtering is an anti-feature. I think that filtering is more clearly implemented as a second step:
```python
    def extract(self, filepath, existing):
        entries = []
        for entry in super().extract(filepath, existing):
            if isinstance(entry, data.Transaction) and entry.postings[0].units.number == ZERO:
                continue
            entries.append(entry)
    return entries        
```
but I may be convinced otherwise.

ThomasdenH:
> I find it much easier to implement the identify() and the archiving related methods directly in the importer. If your experience is different I would like to hear about it.
Maybe some convenient function somewhere to make checking the mime-type doable in one line? Other than that, overriding the methods was indeed easier.

> Do you have in idea for a suitable API?

Your example was pretty much how I handled it. My specific case could be solved with a `decimal_separator=','` in the `csvbase.Amount` constructor, but not sure if that's worth it.

> I think that filtering is more clearly implemented as a second step:
In my case this is a good solution. In the general case there may be situations where the transactions for some rows can't even be built. Is that even possible, if some attributes don't appear for a row? In that case it may be required to filter rows before processing the transactions.

dnicolodi:
> Maybe some convenient function somewhere to make checking the mime-type doable in one line?

Currently it is three lines:
```python
mimetype, encoding = mimetypes.guess_type(filepath)
if mimetype == 'text/csv':
    return True
```
which, loosing some readability, can be reduced to
```python
return mimetypes.guess_type(filepath)[0] == 'text/csv'
```
It would be nicer if `guess_type()` would return a named tuple, but this is not the case, unfortunately.

It could be simplified to something like
```python
return mimetypes.check_type(filepath, 'text/csv')
```
but I don't know if it add enough convenience to make it worth remembering a new API.

> In my case this is a good solution. In the general case there may be situations where the transactions for some rows can't even be built. Is that even possible, if some attributes don't appear for a row? In that case it may be required to filter rows before processing the transactions.

In this case my advice is to override the `etract()` method entirely. It is pretty straightforward. Alternatively, the `read()` method can be overwritten:
```python
    def read(filepath):
        for row in super().read(filepath):
            if not row.amount:
                continue
            yield row
```
This has the drawback that you loose the correct line number information in the transactions metadata but in most applications it does not matter at all (it is not serialized by default and it is there only to provide the correct ordering). I think that in the vast majority of cases a missing filed should result in an error and adding one option to skip rows with missing values would require a way to specify which values are optional, which seems a lot of work for very little gain. The API has been thought too be extensible more than configurable, and I think it works rather well.

dnicolodi:
> Your example was pretty much how I handled it. My specific case could be solved with a `decimal_separator=','` in the `csvbase.Amount` constructor, but not sure if that's worth it.

I added a `subs` argument to the `csvbase.Amount` object constructor. It takes a dictionary mapping regular expression patters to replacements strings. The replacements are applied with `re.sub()` in the order provided before passing the string to `decimal.Decimal()`. I figure that this is the simplest most generic way to handle it. For example, it allows to handle amounts with currency prefixes: `csvbase.Amount(0, {'\\$(.*)', '\\1', ',', '.'})` parses strings like `$1,000.00`.

Thank you very much for your feedback.

ThomasdenH:
Great! Thank you for the work on this project!

PR #99, opened 2022-01-17
Ignore entries without amount in amount_map

### LINKED ISSUE #98: Deduplicator doesn't support entries with price and inferred amount
I have this in my ledger:

```
2016-09-25 * "Hooli Vest Event"
  Assets:US:Schwab:HOOL               1 HOOL {786.9 USD}
  Income:Salary:Hooli:GSU     CHF @ 0.968892 USD
  Assets:US:Hooli:Unvested:C252012  -1 HOOL.UNVEST
  Expenses:Hooli:Vested              1 HOOL.UNVEST
```

Exception:

```
  Traceback (most recent call last):
    File "/home/erik/.local/lib/python3.7/site-packages/beangulp/__init__.py", line 86, in _extract
      entries = extract.extract_from_file(importer, filename, existing_entries)
    File "/home/erik/.local/lib/python3.7/site-packages/beangulp/extract.py", line 44, in extract_from_file
      entries = importer.deduplicate(entries, existing_entries)
    File "/home/erik/.local/lib/python3.7/site-packages/beangulp/importer.py", line 167, in deduplicate
      return extract.mark_duplicate_entries(entries, existing, window, self.cmp)
    File "/home/erik/.local/lib/python3.7/site-packages/beangulp/extract.py", line 125, in mark_duplicate_entries
      if compare(entry, target):
    File "/home/erik/.local/lib/python3.7/site-packages/beangulp/importer.py", line 145, in cmp
      return compare(a, b)
    File "/home/erik/.local/lib/python3.7/site-packages/beangulp/similar.py", line 101, in __call__
      amounts1 = self.cache[id(entry1)] = amounts_map(entry1)
    File "/home/erik/.local/lib/python3.7/site-packages/beangulp/similar.py", line 151, in amounts_map
      amounts[key] += posting.units.number
  TypeError: unsupported operand type(s) for +=: 'decimal.Decimal' and 'type'
```

henrikssn:
Fixes #98

### DIFF
--- a/beangulp/similar.py
+++ b/beangulp/similar.py
@@ -11,6 +11,7 @@ import datetime
 from beancount.core.number import D
 from beancount.core.number import ZERO
 from beancount.core.number import ONE
+from beancount.core.number import MISSING
 from beancount.core import data
 from beancount.core import amount
 from beancount.core import interpolate
@@ -145,7 +146,7 @@ def amounts_map(entry):
         if posting.meta and interpolate.AUTOMATIC_META in posting.meta:
             continue
         currency = isinstance(posting.units, amount.Amount) and posting.units.currency
-        if isinstance(currency, str):
+        if isinstance(currency, str) and posting.units.number != MISSING:
             key = (posting.account, currency)
             amounts[key] += posting.units.number
     return amounts
--- a/beangulp/similar_test.py
+++ b/beangulp/similar_test.py
@@ -113,6 +113,7 @@ class TestDups(cmptest.TestCase):
             2016-01-03 *
               Expenses:Tips         1.01 USD
               Expenses:Tips         1.02 USD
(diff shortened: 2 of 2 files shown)


dnicolodi:
This avoids the exception that is otherwise raised when a `MISSING` value is encountered, however, I suspect that with this change the similarity comparison would be unable to match transactions with posting with missing values. Adding a test case for this would be very helpful.

henrikssn:
> This avoids the exception that is otherwise raised when a `MISSING` value is encountered, however, I suspect that with this change the similarity comparison would be unable to match transactions with posting with missing values. Adding a test case for this would be very helpful.

Yeah, I think you are right... Would a better solution be to store an `(account, key): None` entry in the map?

PR #105, opened 2022-08-23
Adds CSV MIME type that compatable with Windows

Gateswong:
Mentioned by https://github.com/jbms/beancount-import/issues/169.

### DIFF
--- a/beangulp/importers/csv.py
+++ b/beangulp/importers/csv.py
@@ -164,7 +164,7 @@ def prepare_for_identifier(regexps: Union[str, List[str]],
     if isinstance(regexps, str):
         regexps = [regexps]
     matchers = matchers or []
-    matchers.append(('mime', 'text/csv'))
+    matchers.append(('mime', '(text/csv)|(application/vnd\\.ms-excel)'))
     if regexps:
         for regexp in regexps:
             matchers.append(('content', regexp))


blais:
So this mentions excel... not CSV. Why?

hexhexD:
I think having Excel as the default app for CSV results a MIME of application/vnd.ms-excel (It doesn't happen using other program as default app for CSV extension). I tested this patch and it can recognize CSV with or without Excel installation so I think it's a improvement over the original.

beangulp/importers/csv.py
```diff
@@ -164,7 +164,7 @@ def prepare_for_identifier(regexps: Union[str, List[str]],
     if isinstance(regexps, str):
         regexps = [regexps]
     matchers = matchers or []
-    matchers.append(('mime', 'text/csv'))
+    matchers.append(('mime', '(text/csv)|(application/vnd\\.ms-excel)'))
```
dnicolodi, line 167: The problem with this change is that the importers based on `beangulp.importers.csv` will start matching all files with the Excel mime type, including xls ans xlsx files. This would be a backward compatibility break, and we should avoid this, especially for an importer base class that has been around for very very long.

What can be done is to add the `mime` matcher only if one has not been passed by the user already. This would allow you to add the matching of the `application/vnd\\.ms-excel` to your importer only. Something like this:
```suggestion
    if not any(m[0] == 'mime' for m in matchers):
        matchers.append(('mime', 'text/csv'))
```
if I remember correctly how this code is supposed to work.

dnicolodi (changes requested):

PR #109, opened 2023-02-16
Add Option to Set Date Separator During Archive

patbakdev:
I prefer to use a space between the date and original filename.

### DIFF
--- a/beangulp/__init__.py
+++ b/beangulp/__init__.py
@@ -116,8 +116,10 @@ def _extract(ctx, src, output, existing, reverse, failfast, quiet):
               help='Just print where the files would be moved.')
 @click.option('--failfast', '-x', is_flag=True,
               help='Stop processing at the first error.')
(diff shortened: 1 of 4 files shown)


dnicolodi:
The reason for having the filename structured in the current way is because beancounts automatically picks up files with names in that format to add `document` directives to the ledger. Changing the separator between the date and the rest of the name would prevent this mechanism from working.

I'm not opposed to make it configurable (and a space looks better than a dot to me too) but I don't think it should be a command line parameter for two reasons: the users are going to get inconsistent file names if they forgot the option, and as explained above, because it does not mix well with the support for automatic `document` directives.

What I suggest to do, is to use a module global variable for the filename template. I'll prepare a PR.

dnicolodi:
I've just pushed the implementation. The solution is to add
```python
import beangulp
beangulp.archive.FILENAME = '{date:%Y-%m-%d} {name}'
```
or any other format you like, to your import script.

PR #110, opened 2023-02-16
Add Option to Strip Existing Date During Archive

patbakdev:
Sometime I save file with an initial save date, but the transaction date isn't for a few days later. This makes it easy to update the file instead of having to deal with the error.

### DIFF
--- a/beangulp/__init__.py
+++ b/beangulp/__init__.py
@@ -116,8 +116,10 @@ def _extract(ctx, src, output, existing, reverse, failfast, quiet):
               help='Just print where the files would be moved.')
 @click.option('--failfast', '-x', is_flag=True,
               help='Stop processing at the first error.')
+@click.option('--strip-date', '-sd', is_flag=True,
+              help='Remove existing date from filename if it exists.')
 @click.pass_obj
-def _archive(ctx, src, destination, dry_run, overwrite, failfast):
+def _archive(ctx, src, destination, dry_run, overwrite, failfast, strip_date):
     """Archive documents.
 
     Walk the SRC list of files or directories and move each file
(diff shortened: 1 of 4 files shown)


dnicolodi:
I'm not sure I understand the purpose. If you want to rename the archived files, why don't you implement the `filename()` method for your importer? I'm extremely reluctant to add more options to the basic command line tools, especially very use-case specific options.

patbakdev:
Ignore my original comment above, got my wires crossed. 

Basically sometimes when I download a file it contains date and I have to manually rename it or I will run into this error when I try and archive the file.

```
if re.match(r'\d\d\d\d-\d\d-\d\d\.', filename):
    raise Error("The filename contains what looks like a date.")
```

This change will allow me to strip that date out automatically. Another options is just to always strip out the date, but since I was already making changes that needed options, it seemed safest to surface as on option and maintain the existing default behavior.

As for the options. I was originally going to provide a global dictionary for options to be defined int the configuration file, but decided this would be better since it was more in line with the existing interface. It would have looked something like this.

```
    def __init__(self, importers, hooks=None, options=None):
        self.importers = [_importer(i) for i in importers]
        self.hooks = list(hooks) if hooks is not None else []
        self.options=options
```

I have another PR that is almost ready to print all entries in date order (no sections), plus #109. I could move them all to using config based options instead. I'm just not sure what that data structure should look like. I loose dictionary or class/obj.

dnicolodi:
You get that error only if the `filename()` method of your importer class returns a file name that contains a date. Just don't do that: strip the date in the `filename()` method of your importer. No need to add this support for stripping dates to the archive command nor to add an option for it.

PR #111, opened 2023-02-16
Add Option to Skip Sections During Extract

patbakdev:
I prefer to have a single output file with all transactions (entities) sorted by date instead of the default behavior to add a section for each imported file. This option will accomplish that.

### DIFF
--- a/beangulp/__init__.py
+++ b/beangulp/__init__.py
@@ -49,12 +49,14 @@ def _walk(file_or_dirs, log):
               help='Existing Beancount ledger for de-duplication.')
 @click.option('--reverse', '-r', is_flag=True,
               help='Sort entries in reverse order.')
+@click.option('--no-sections', '-ns', is_flag=True,
+              help='Separate files into sections')
 @click.option('--failfast', '-x', is_flag=True,
               help='Stop processing at the first error.')
 @click.option('--quiet', '-q', count=True,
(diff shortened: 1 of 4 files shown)


dnicolodi:
Sections are there to allow to easily make sense of the output, especially in presence of duplicates (the order in which the input files are processed matters). If you don't like the section, just post-process the output to remove them:
```python
import click
from beancount.parser import parser
from beancount.parser import printer

@click.command()
@click.argument('ledger')
def main(ledger):
    entries, errors, options = parser.parse_file(ledger)
    printer.print_entries(entries)

if __name__ == '__main__':
    main()
```

patbakdev:
Thanks. I didn't think re-parse it. I'll give that a go.

dnicolodi:
I much prefer composability to adding many options to tools. Since a long while we have an idea for a `bean-gadget` tool that implements this sort of simple command line tools, but no one has found the time to put it together yet.

patbakdev:
Makes sense. I am basically building my own personal version of "bean-gadget" for my purposes with commands to download, commit/backup, import (calls out to beangulp), and export (custom exporters for tax filing, reporting, etc.) As part of that I had hacked together a bench of changes to beangulp as I went along. Had a few hours available so I thought I would clean it up and see if it was of value to the wider groups. 😄

PR #118, opened 2023-08-09
Add pytest hooks for importer integration tests

ericaltendorf:
Migrated with adaptations from v2 codebase.

### DIFF
--- a/beangulp/regression_pytest.py
+++ b/beangulp/regression_pytest.py
@@ -1,0 +1,181 @@
+"""Support for implementing regression tests on sample files using pytest.
+
+This module provides definitions for testing a custom importer against a set of
+existing downloaded files, running the various importer interface methods on it,
+and comparing the output to an expected text file. (Expected test files can be
+auto-generated using the --generate option). You use it like this:
+
+  from beangulp import regression_pytest
+  ...
+  import mymodule
+  ...
+
+  # Create your importer instance used for testing.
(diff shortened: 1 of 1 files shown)


Goorzhel:
This patch helped me get unstuck on [making my regression tests work](https://git.sr.ht/~goorzhel/beancount-clipper/commit/af18e9925f52831231090f369c2915ec28781144) on Beangulp.

johannesjh:
I like the idea of exposing importer tests as pytest/unittest testcases. Regarding the implementation: I think parts of what this pull request contributes exists already in beangulp:

What exists already:
- a CLI for running such regression tests, see `beangulp.testing.main`
- the logic for these regression tests, compare `beangulp.testing._test`
- the logic for reading input and expected output, see `beangulp.testing._run`

What does not exist yet, as far as I see: 
- a simple way of exposing beangulp's regressions test functionality to testframeworks such as `unittest` or `pytest`. That's what I like about this pull request, I would like that, too.

Question / Suggestion for this pull request: 
Would it be better for this pull request to use more of the existing functionality?

dnicolodi:
> Would it be better for this pull request to use more of the existing functionality?

Most definitely. This is the main reason why this stalled. Furthermore, it should also expose the functionality in a way useful also for integration with an `unittest` test suite, not only with `pytest`. Unfortunately, I do not have time to work on this myself or to provide detailed guidance on how I think this would be best implemented.

PR #125, opened 2024-06-28
Fix MIME type check for example OFX importer.

josephw:
Ensure that we will accept the non-strict type we added, and don't try to match a tuple against a string. Without this, type check will always fail, and no files will pass identification.


beangulp/__init__.py
```diff
@@ -188,8 +188,8 @@ def _archive(ctx, src, destination, dry_run, overwrite, failfast):
         sys.exit(1)
 
     if not dry_run:
-        for src, dst in renames:
```
josephw, line 191: Avoid failing linting.

PR #126, opened 2024-07-06
Support CSV files where the amount isn't in a single column and the value embeds currency symbol

rajive:
Support use cases where the CSV file doesn't have amount in a single column---instead the amount is derived from multiple columns. In my case (see unit test case [test_extract_amount](https://github.com/rajive/beangulp/blob/0892d2c73364a5e21a128d04d83deae27b72203c/beangulp/importers/csvbase_test.py#L424)), the bank's CSV had a "Withdrawal" column and a "Deposit" column. In each row, only one on those columns has a positive numeric value. 

Furthermore, the numeric value embeds currency ('$') symbol with comma separated digits that were not being parsed correctly.

### DIFF
--- a/beangulp/importers/csvbase.py
+++ b/beangulp/importers/csvbase.py
@@ -262,8 +262,13 @@ class Importer(beangulp.Importer, CSVReader):
         """Implement beangulp.Importer::extract()
 
         This methods costructs a transaction for each data row using
-        the date, narration, and amount required fields and the flag,
+        the date, narration required fields and the flag,
         payee, account, currency, tag, link, balance optional fields.
+
+        The amount() method returns the amount for a row---the default
+        implementation requires an amount field. The method can be
+        redefined in subclasses---for use cases where the amount is
+        derived from multiple columns, e.g. deposit and withdrawal.
 
         Transaction metadata is constructed with the metadata() method
         and the finalize() method is called on each transaction. These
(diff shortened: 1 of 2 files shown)


dnicolodi:
Thanks for your contribution. I haven't had a close look yet. However, the idea behind the `csvbase` module is to provide an extensible base for writing importers dealing with CSV files, not to provide configuration knobs for every possible case. The latter causes combinatorial explosion of complexity and configurations that need to be tested, which is the main motivation for replacing `beanqulp.importers.csv` with `beangulp.importers.csvbase`. Therefore, if it exists a way to solve the problem that avoid adding complexity to the base class, it should be preferred to extending the base classes. I'll have a closer look and I'll let you know.

dnicolodi:
This should work for your use case:
```python
from beangulp.importers import csvbase
from beangulp.utils import parse_amount

class WithdrawalOrDeposit(csvbase.Column):
    def __init__(self, withdrawal, deposit):
        super().__init__(withdrawal, deposit)

    def parse(self, withdrawal, deposit):
        if withdrawal and deposit:
            raise ValueError
        return -parse_amount(withdrawal) if withdrawal else parse_amount(deposit)


    date = csvbase.Date('Date', '%m/%d/%Y')
    narration = csvbase.Columns('Description')
    amount = WithdrawalOrDeposit('Withdrawal', 'Deposit')
```

dnicolodi:
No contributor responses to review in a few months. Closing

PR #129, opened 2024-10-22
archive: Add --quiet option

jakobbjorns:
I'm using a shell script to automatically extract and archive files , and I'm using the -q flag in `extract` to suppress unnecessary printing. I realized that a such flag doesn't exist for `archive`. I implemented it by copying code from the `_extract`-function to the `_archive`-function

### DIFF
--- a/beangulp/__init__.py
+++ b/beangulp/__init__.py
@@ -52,7 +52,7 @@ def _walk(file_or_dirs, log):
 @click.option('--failfast', '-x', is_flag=True,
               help='Stop processing at the first error.')
 @click.option('--quiet', '-q', count=True,
-              help="Suppress all output.")
+              help='Suppress all output.')
 @click.pass_obj
 def _extract(ctx, src, output, existing, reverse, failfast, quiet):
     """Extract transactions from documents.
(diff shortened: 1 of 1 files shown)


dnicolodi (changes requested):
Can you please fix the missing space and reword the commit message to something like
`archive: Add --quiet option`

beangulp/__init__.py
```diff
@@ -146,8 +148,9 @@ def _archive(ctx, src, destination, dry_run, overwrite, failfast):
     if destination is None:
         import __main__
         destination = os.path.dirname(os.path.abspath(__main__.__file__))
-
-    log = utils.logger()
+
```
dnicolodi, line 151:
```suggestion

```
There is some extra whitespace that needs to be removed here.

beangulp/__init__.py
```diff
@@ -122,8 +122,10 @@ def _extract(ctx, src, output, existing, reverse, failfast, quiet):
               help='Just print where the files would be moved.')
 @click.option('--failfast', '-x', is_flag=True,
               help='Stop processing at the first error.')
+@click.option('--quiet', '-q', count=True,
+              help="Suppress all output.")
```
dnicolodi, line 126:
```suggestion
              help='Suppress all output.')
```
The code you copied also uses the wrong quote style. You can add a patch to fix that code as well, if you want.

jakobbjorns:
Thanks for your feedback! I fixed the comments you had. I'm quite new to git, I hope i did it the right way

dnicolodi (approved):

dnicolodi:
Thanks for your contribution!

PR #135, opened 2024-12-17
docs: improve example by providing TL;DR

Sieboldianus:
I tried to reproduce the example but stumbled with the currently available documentation. Perhaps the example can be split into a separate repository with a minimal TL;DR? Anyway, I added the minimal necessary steps to reproduce the example in Linux to the example `Readme.md`. Maybe this helps some new people.


blais:
Thank you

PR #136, opened 2024-12-22
refactor: add idify

trim21:
https://github.com/beancount/beancount/pull/907#issuecomment-2558518208


beangulp/importer.py
```diff
@@ -11,7 +11,7 @@
 
 from beancount.core import flags
 from beancount.core import data
-from beancount.utils import misc_utils
+from beancount.utils import idify
```
blais, line 14: I think you meant beangulp here.

PR #140, opened 2024-12-28
Fix keyword args in legacy CSV Importer

jeremander:
The named keyword arguments in the (v2 legacy) `beangulp.importers.csv.Importer` do not seem to get passed to the parent class in the `__init__` method.  This can lead to unexpected behavior (for instance, I set `invert_sign` to `True`, but it was silently ignored).  To fix this, we supply the keyword arguments to the call to `super().__init__`.


dnicolodi:
Thanks! I edited the commit message slightly.

PR #141, opened 2025-01-02
add py.typed marker

yagebu:


yagebu:
Build failures are unrelated, see #142 for a PR that fixes them.

Tested locally by installing it, the py.typed marker was installed as well.

dnicolodi:
There isn't much value in declaring the package has having typing annotations without verifying that they are complete and correct. It would be counterproductive, actually. Given that the recent work on adding typing to Beancount has created more issue than what it solved, I really don't want to replicate the experience for beangulp users.

blais:
@dnicolodi Why didn't we modernize all the projects to more recent requirements? 3.10 and above. Gotta let go of the past at some point

yagebu:
> There isn't much value in declaring the package has having typing annotations without verifying that they are complete and correct. It would be counterproductive, actually.

I disagree. I find it very helpful to have working types for base classes like Importer already even if not all functions and modules are fully typed. I think the incremental approach suggested by mypy (https://mypy.readthedocs.io/en/stable/existing_code.html) works on an ecosystem level as well. Currently, to get type-checking for the bits that are used in Fava I maintain type stubs - those could be removed and all beangulp users could profit from the type annotations.

yagebu:
I can also create a PR to run mypy in CI so that the types are checked. There's only a handful of errors

dnicolodi:
I think you misunderstood my reply. Typing annotations are useful. Wrong typing annotations are not. A `py.typed` marker indicates that typing information has at least a good chance to be correct. Because no effort has been put into verifying that typing for beangulp is correct, I'm sure it is not. Thus adding a `py.typed` marker seems premature. The process should be to verify that typing is correct and then add a `py.typed` marker. Not the other way around.

If you want to use the existing typing annotations, you can instruct mypy to do it:
```toml
[[tool.mypy.overrides]]
module = ["beangulp.*"]
follow_untyped_imports = true
```

yagebu:
See #143 which adds a CI run of mypy to ensure beangulp has a consistent set of type annotations

yagebu:
> If you want to use the existing typing annotations, you can instruct mypy to do it:

Thanks, TIL - didn't know of this brand-new 1.14 mypy feature yet, that's very useful.

PR #143, opened 2025-01-03
types

yagebu:
- **adjust types for a successful mypy run**
- **ci: add mypy ci step**

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -14,3 +14,14 @@ jobs:
           python-version: '3.11'
       - run: pip install ruff
       - run: ruff check beangulp/ examples/
+  mypy:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.9'
+      - run: pip install . mypy types-python-dateutil types-beautifulsoup4 beancount@git+https://github.com/beancount/beancount.git
+      - run: mypy beangulp examples
--- a/beangulp/__init__.py
+++ b/beangulp/__init__.py
@@ -14,6 +14,7 @@ import os
 import sys
 import warnings
 import click
+from typing import Any, Optional, Union, Sequence
 
 from beancount import loader
 
@@ -257,7 +258,7 @@ def _importer(importer):
 
 
 class Ingest:
-    def __init__(self, importers, hooks=None):
+    def __init__(self, importers: Sequence[Union[Importer, ImporterProtocol]], hooks: Optional[Sequence[Any]] = None) -> None:
         self.importers = [_importer(i) for i in importers]
         self.hooks = list(hooks) if hooks is not None else []
 
--- a/beangulp/__init__.py
+++ b/beangulp/__init__.py
@@ -14,7 +14,7 @@ import os
 import sys
 import warnings
 import click
-from typing import Any, Optional, Union, Sequence
+from typing import Callable, Optional, Union, Sequence
 
 from beancount import loader
 
@@ -258,7 +258,9 @@ def _importer(importer):
 
 
 class Ingest:
-    def __init__(self, importers: Sequence[Union[Importer, ImporterProtocol]], hooks: Optional[Sequence[Any]] = None) -> None:
+    def __init__(self,
+                 importers: Sequence[Union[Importer, ImporterProtocol]],
+                 hooks: Optional[Sequence[Callable]] = None) -> None:
         self.importers = [_importer(i) for i in importers]
         self.hooks = list(hooks) if hooks is not None else []
 
--- a/beangulp/cache.py
+++ b/beangulp/cache.py
@@ -155,7 +155,7 @@ def get_file(filename):
     return _CACHE[filename]
 
 
-_CACHE = utils.DefaultDictWithKey(_FileMemo)
+_CACHE = utils.DefaultDictWithKey(_FileMemo)  # type: ignore
 
 
 def cache(func=None, *, key=None):
--- a/beangulp/importer.py
+++ b/beangulp/importer.py
@@ -177,7 +177,7 @@ class ImporterProtocol:
     # you prefer to create your imported transactions with a different flag.
     FLAG = flags.FLAG_OKAY
 
-    def name(self):
+    def name(self) -> str:
         """See Importer class name property."""
         return f"{self.__class__.__module__}.{self.__class__.__name__}"
 
@@ -185,9 +185,15 @@ class ImporterProtocol:
 
     def identify(self, file) -> bool:
         """See Importer class identify() method."""
+        # Type error ignore for backwards compatibility - this should be overridden
+        # and implemented in subclasses
+        return None  # type: ignore
 
     def file_account(self, file) -> data.Account:
         """See Importer class account() method."""
+        # Type error ignore for backwards compatibility - this should be overridden
+        # and implemented in subclasses
+        return None  # type: ignore
 
     def file_date(self, file) -> Optional[datetime.date]:
         """See Importer class date() method."""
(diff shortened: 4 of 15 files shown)


.github/workflows/lint.yaml
```diff
@@ -14,3 +14,14 @@ jobs:
...
       - run: ruff check beangulp/ examples/
+  mypy:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.9'
```
dnicolodi, line 25: Why Python 3.9?

beangulp/__init__.py
```diff
@@ -257,7 +258,7 @@ def _importer(importer):
 
 
 class Ingest:
-    def __init__(self, importers, hooks=None):
+    def __init__(self, importers: Sequence[Union[Importer, ImporterProtocol]], hooks: Optional[Sequence[Any]] = None) -> None:
```
dnicolodi, line 261: `hooks: Optional[Sequence[Any]]` could at least be `hooks: Optional[Sequence[Callable]]`

beangulp/cache.py
```diff
@@ -155,7 +155,7 @@ def get_file(filename):
     return _CACHE[filename]
 
 
-_CACHE = utils.DefaultDictWithKey(_FileMemo)
+_CACHE = utils.DefaultDictWithKey(_FileMemo)  # type: ignore
```
dnicolodi, line 158: Why is the ignore needed?

beangulp/importer.py
```diff
@@ -177,37 +177,44 @@ class ImporterProtocol:
...
         """See Importer class name property."""
         return f"{self.__class__.__module__}.{self.__class__.__name__}"
 
     __str__ = name
 
     def identify(self, file) -> bool:
         """See Importer class identify() method."""
+        # Type error ignore for backwards compatibility - this should be overridden
+        # and implemented in subclasses
+        return None  # type: ignore
```
dnicolodi, line 190: If returning `None` is not allowed, `return None` should be `raise NotImplementedError`. If it is allowed, the annotation is wrong.

beangulp/importer.py
```diff
@@ -177,37 +177,44 @@ class ImporterProtocol:
...
 
     def file_account(self, file) -> data.Account:
         """See Importer class account() method."""
+        # Type error ignore for backwards compatibility - this should be overridden
+        # and implemented in subclasses
+        return None  # type: ignore
```
dnicolodi, line 196: Same.

beangulp/petl_utils.py
```diff
@@ -1,11 +1,11 @@
 """Utilities using petl.
 """
 
-from typing import Optional
+from typing import Optional, Set
 import datetime
 import re
 
-import petl
+import petl  # type: ignore
```
dnicolodi, line 8: Why this ignore?

beangulp/petl_utils_test.py
```diff
@@ -1,7 +1,7 @@
 import datetime
 import decimal
 import unittest
-import petl
+import petl  # type: ignore
```
dnicolodi, line 4: Why?

beangulp/testing.py
```diff
@@ -46,7 +46,7 @@ def write_expected_file(filepath: str, *data, force: bool = False):
     """Writes out the expected file."""
     mode = 'w' if force else 'x'
     with open(filepath, mode) as expfile:
-        write_expected(expfile, *data)
+        write_expected(expfile, *data)  # type: ignore
```
dnicolodi, line 49: Why?

examples/importers/csvbank.py
```diff
@@ -5,7 +5,7 @@
 
 
 class Importer(csvbase.Importer):
-    date = csvbase.Date('Posting Date', '%m/%d/%Y')
+    date = csvbase.Date('Posting Date', '%m/%d/%Y')  # type: ignore
```
dnicolodi, line 8: Why?

examples/importers/utrade.py
```diff
@@ -52,27 +52,28 @@ def identify(self, filepath):
...
-        entries = []
+        entries: data.Entries = []
         index = 0
         with open(filepath) as infile:
             for index, row in enumerate(csv.DictReader(infile)):
                 meta = data.new_metadata(filepath, index)
                 date = parse(row['DATE']).date()
                 rtype = row['TYPE']
                 link = f"ut{row['REF #']}"
+                links = frozenset([link])
```
dnicolodi, line 76: These two lines can be one line.

examples/importers/utrade.py
```diff
@@ -122,9 +123,9 @@ def extract(self, filepath, existing):
                     rate = D(match.group(3))
 
                     if rtype == 'BUY':
-                        cost = position.Cost(rate, self.currency, None, None)
+                        cost = position.CostSpec(rate, None, self.currency, None, None, None)
```
dnicolodi, line 126: Why? `Cost` should be just fine.

examples/importers/utrade.py
```diff
@@ -143,11 +144,11 @@ def extract(self, filepath, existing):
                             logging.error("Missing cost basis in '%s'", row['DESCRIPTION'])
                             continue
                         cost_number = D(match.group(1))
-                        cost = position.Cost(cost_number, self.currency, None, None)
+                        cost = position.CostSpec(cost_number, None, self.currency, None, None, None)
```
dnicolodi, line 147: Here too.

beangulp/petl_utils.py
```diff
@@ -47,32 +47,33 @@ def table_to_directives(
...
         units = amount.Amount(rec.amount, currency)
-        tags, links = set(), set()
+        tags: Set[str] = set()
+        links: Set[str] = set()
+        link = getattr(rec, "link", None)
+        if link:
+            links.add(link)
+        tag = getattr(rec, "tag", None)
+        if tag:
+            tags.add(tag)
```
dnicolodi, line 62: I'm not entirely convinced that the type for `links` and `tags` should be `frozenset` and not simply `collections.abc.Set`. Anyhow,  This seems a very convoluted way to crete a `frozenset` of one element.

dnicolodi (changes requested):
I'm not convinced that adding types to the examples is a good idea. The examples are supposed to be as simple as possible and serve as templates for inexperienced programmers. Adding typing annotations seems to only add complexity for very little gain.

.github/workflows/lint.yaml
yagebu: that's the oldest supported version of Beancount - which version do you want?

beangulp/importer.py
yagebu: sounds good, will change that - wasn't sure whether you'd want this sort of breaking change

examples/importers/utrade.py
yagebu: a Cost requires a date (the booking code will insert it for all postings, falling back to the entry date)

beangulp/petl_utils.py
yagebu: didn't look much further into it - is this file even supposed to be part of the package? petl doesn't seem to be listed in the dependencies

beangulp/petl_utils.py
yagebu: I stuck to the original code but I'm happy to refactor this as well

examples/importers/csvbank.py
yagebu: This date column shadows the date function of the importer protocol. Seems like the base class initialisation makes that all work but I don't know if this can be typed correctly.

beangulp/testing.py
yagebu: expfile is `IO[Any]` and write_expected takes `TextIO` - not really sure why mypy doesn't infer that it's in text mode

.github/workflows/lint.yaml
Goorzhel: If it helps to narrow down the choices: 3.9 is the oldest version with security-only support, but 3.12 is the oldest with [full support](https://peps.python.org/pep-0693/).

yagebu:
Closing, there's quite some merge conflicts and I think generally some disagreement about the usefulness of this

PR #144, opened 2025-01-03
make 'finialize' can remove current record

suqin-haha:
it helps to exclude some records. 

Please let me know if there is a better way to do that. :D


dnicolodi:
Thanks. It looks like an useful addition. I changed the approach slightly, adjusted the docstrings, squashed the commits, and adjusted the commit message. Please have a look.

Issue #145, opened 2025-01-04
Balance wrong for csvbase importers with multiple transactions in a day


rbrunt:
I've just migrated over to beancount3, and while doing so I decided to move my importers over to use the new `csvbase.Importer` base class since the older csv base class has now been marked as deprecated.

Overall, this has been a great change - it's now much simpler to understand how to implement them, and they're much smaller now, so easier to understand what's going on.

The only issue I've encountered is that the automatic balance assertions always fail when there are multiple transactions on the last day of the csv's listings, since it takes the first entry in the last day. All of the csv files I have from my bank are in ascending date order, so this means that the balance is always off by the amount of the last transaction.

I've got a new (failing) test that reproduces the issue to communicate what it is that I mean here: https://github.com/rbrunt/beangulp/commit/528deba85236be8209d380a6d1c9b4675cac01b9

```python
    @docfile
    def test_extract_balance_with_multiple_transactions_per_day_and_ascending_dates(self, filename):
        """\
        2021-05-17, Test A, 1.00, 1.00
        2021-05-18, Test B, 2.00, 3.00
        2021-05-18, Test C, 1.00, 4.00
        """
        class CSVImporter(Base):
            date = Date(0)
            narration = Column(1)
            amount = Amount(2)
            balance = Amount(3)
            names = False
        importer = CSVImporter('Assets:CSV', 'EUR')
        entries = importer.extract(filename, [])
        self.assertEqualEntries(entries, """
        2021-05-17 * "Test A"
          Assets:CSV  1.00 EUR
        2021-05-18 * "Test B"
          Assets:CSV  2.00 EUR
        2021-05-18 * "Test C"
          Assets:CSV  1.00 EUR
        2021-05-19 balance Assets:CSV  4.00 EUR
        """)
```

I'm happy to write a tweak to `csvbase.Importer` to implement this behaviour how I would have expected it to work (to get this new test to pass), but before I do, I wanted to check if there were any views on approach: I don't want to change behaviour here and affect others who update to a new version because the base class has changed under them, but I also think this is likely to be catching quite a few people out.

The easiest approach would be to move from using `max` to a more explicit pick:

```python
for currency, balances in balances.items():
            ascending = balances[0].date >= balances[-1].date
            last_balance = balances[0] if ascending else balances[-1]
            entries.append(last_balance)
```

However this might break existing importers people have already written, so perhaps pulling this bit of code out into a separate function that can be overriden by a subclass or mixin would be a safer bet.

[...]

PR #146, opened 2025-01-06
Fix wrong balance being added for csvbase importers when csv file is in date ascending order

rbrunt:
This PR fixes a bug where the the wrong balance assertion was picked for CSV with entries in date ascending order. For a little more context, see https://github.com/beancount/beangulp/issues/145.

I've added a test to catch future regressions.

### DIFF
--- a/beangulp/importers/csvbase.py
+++ b/beangulp/importers/csvbase.py
@@ -2,6 +2,7 @@ import abc
 import csv
 import datetime
 import decimal
+from enum import Enum
 import re
 
 from collections import defaultdict
@@ -168,17 +169,26 @@ class CSVMeta(abc.ABCMeta):
         return super().__new__(mcs, name, bases, others)
 
 
+class Order(Enum):
+    ASCENDING = 1
+    """Entries are listed in chronological order."""
+    DESCENDING = 2
+    """Entries are listed in reverse chronological order."""
+
+
 class CSVReader(metaclass=CSVMeta):
     encoding = 'utf8'
     """File encoding."""
     skiplines = 0
-    """Number of input lines to skip before startign processing."""
+    """Number of input lines to skip before starting processing."""
     names = True
     """Whether the data file contains a row with column names."""
     dialect = None
     """The CSV dialect used in the input file."""
     comments = '#'
     """Comment character."""
+    order = None
+    """Order of entries in the CSV file. If None the order will be inferred from the file content."""
 
     # This is populated by the CSVMeta metaclass.
     columns = {}
@@ -323,13 +333,16 @@ class Importer(beangulp.Importer, CSVReader):
         if not entries:
             return []
 
+        if self.order is None:
+            self.order = Order.ASCENDING if entries[0].date <= entries[-1].date else Order.DESCENDING
+
         # Reverse the list if the file is in descending order.
-        if not entries[0].date <= entries[-1].date:
+        if self.order is Order.DESCENDING:
             entries.reverse()
 
         # Append balances.
         for currency, balances in balances.items():
-            entries.append(max(balances, key=lambda x: x.date))
+            entries.append(balances[-1 if self.order is Order.ASCENDING else 0])
 
         return entries
 
--- a/beangulp/importers/csvbase_test.py
+++ b/beangulp/importers/csvbase_test.py
@@ -5,7 +5,7 @@ import unittest
 from beancount.core import data
 from beancount.parser import cmptest
 from beancount.utils.test_utils import docfile
-from beangulp.importers.csvbase import Column, Columns, Date, Amount, CSVMeta, CSVReader, Importer
+from beangulp.importers.csvbase import Column, Columns, Date, Amount, CSVMeta, CSVReader, Order, Importer
 
 
 class TestColumn(unittest.TestCase):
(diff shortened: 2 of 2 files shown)


dnicolodi:
I don't think this is the right approach. This still gives you the wrong result if there source has data relative to only one day.

rbrunt:
> I don't think this is the right approach. This still gives you the wrong result if there source has data relative to only one day.

@dnicolodi Sorry, I don't think I've understood what you mean. Could you give an example input?

E.g. the following input

```csv
2021-05-18, Test B, 2.00, 3.00
2021-05-18, Test C, 1.00, 4.00
```

Does give the correct output:

```beancount
 2021-05-18 * "Test B"
          Assets:CSV  2.00 EUR
2021-05-18 * "Test C"
          Assets:CSV  1.00 EUR
2021-05-19 balance Assets:CSV  4.00 EUR
```

... though having written that I realise that I'm assuming that we should be reading down the file here. Is that what you mean?

dnicolodi:
That is because the entries in the file are sorted in execution order. It does not work if the entries are sorted in reversed execution order: the check that tries to detect this occurrence does not work if there is data relative to just one day.

rbrunt:
Yes, I see.

At the moment balances are not reliable for CSVs in execution order. This patch fixes that as well as the edge case where they're in exeuction order and there's only one date listed, but doesn't catch the edge case where they're in reverse execution order and there's only one date listed.

I suppose really the general solution is to not try to guess the ordering at all and force the inheritor to specify - how can we assume what order the entries are in if there's only one day? That's a breaking change though, and I'm not sure how often a CSV with only one entry would appear...

I'm still quite new to beancount, so I'm quite aware that there might be assumptions about how things should behave that I'm not familiar with. Very happy to have an alternative approach, as long as the bug with balance assertions for csvs in execution order is fixed.

blais:
I think to handle that corner case (rare I suppose) you could add an argument that sets the expected ordering instead of inferring it from the data. That would be a reasonable thing to do IMHO.

blais:
(Approved to run checks.)

beangulp/importers/csvbase.py
```diff
@@ -168,17 +169,28 @@ def __new__(mcs, name, bases, dct):
         return super().__new__(mcs, name, bases, others)
 
 
+class EntryOrder(Enum):
+    Infer = 0
+    ExecutionOrder = 1
+    """Entries are listed from first to last"""
+    ReverseExecutionOrder = 2
+    """Entries are listed from last to first"""
```
dnicolodi, line 177:
```suggestion
class Order(Enum):
    ASCENDING = 1
    """Entries are listed in chronological order."""
    DESCENDING = 2
    """Entries are listed in reverse chronological order."""
```

beangulp/importers/csvbase.py
```diff
@@ -323,13 +335,19 @@ def extract(self, filepath, existing):
         if not entries:
             return []
 
+        entries_in_execution_order = entries[0].date <= entries[-1].date
+        if (self.entry_order != EntryOrder.Infer):
+            entries_in_execution_order = self.entry_order == EntryOrder.ExecutionOrder
+
```
dnicolodi, line 341:
```suggestion
        if self.order is None:
            self.order = Order.ASCENDING if entries[0].date <= entries[-1].date else Order.DESCENDING
```

beangulp/importers/csvbase.py
```diff
@@ -168,17 +169,28 @@ def __new__(mcs, name, bases, dct):
...
+    """Number of input lines to skip before starting processing."""
     names = True
     """Whether the data file contains a row with column names."""
     dialect = None
     """The CSV dialect used in the input file."""
     comments = '#'
     """Comment character."""
+    entry_order = EntryOrder.Infer
+    """Order of entries in the csv file. Default is 'Infer', where the reader will try to guess.
+    If the guess isn't correct, you can override by passing a different value."""
```
dnicolodi, line 193:
```suggestion
    order = None
    """Order of entries in the CSV file. If None the order will be inferred from the file content."""
```

beangulp/importers/csvbase.py
dnicolodi, line 341:
```suggestion
        if self.order is None:
            self.order = Order.ASCENDING if entries[0].date <= entries[-1].date else Order.DESCENDING
```

beangulp/importers/csvbase.py
```diff
@@ -323,13 +335,19 @@ def extract(self, filepath, existing):
         if not entries:
             return []
 
+        entries_in_execution_order = entries[0].date <= entries[-1].date
+        if (self.entry_order != EntryOrder.Infer):
+            entries_in_execution_order = self.entry_order == EntryOrder.ExecutionOrder
+
         # Reverse the list if the file is in descending order.
-        if not entries[0].date <= entries[-1].date:
+        if not entries_in_execution_order:
```
dnicolodi, line 343:
```suggestion
        if self.order is Order.DESCENDING:
```

beangulp/importers/csvbase.py
```diff
@@ -323,13 +335,19 @@ def extract(self, filepath, existing):
...
             entries.reverse()
+            for balance_list in balances.values():
+                balance_list.reverse()
```
dnicolodi, line 346:
```suggestion
```

beangulp/importers/csvbase.py
```diff
@@ -323,13 +335,19 @@ def extract(self, filepath, existing):
...
 
         # Append balances.
         for currency, balances in balances.items():
-            entries.append(max(balances, key=lambda x: x.date))
+            entries.append(balances[-1])
```
dnicolodi, line 350:
```suggestion
            entries.append(balances[-1 if self.order is Order.ASCENDING else 0])
```

beangulp/importers/csvbase_test.py
```diff
@@ -5,7 +5,7 @@
 from beancount.core import data
 from beancount.parser import cmptest
 from beancount.utils.test_utils import docfile
-from beangulp.importers.csvbase import Column, Columns, Date, Amount, CSVMeta, CSVReader, Importer
+from beangulp.importers.csvbase import Column, Columns, Date, Amount, CSVMeta, CSVReader, EntryOrder, Importer
```
dnicolodi, line 8:
```suggestion
from beangulp.importers.csvbase import Column, Columns, Date, Amount, CSVMeta, CSVReader, Order, Importer
```

beangulp/importers/csvbase_test.py
```diff
@@ -356,6 +356,54 @@ class CSVImporter(Base):
...
+        2021-05-18, Test B, 2.00, 4.00
+        2021-05-18, Test C, 1.00, 2.00
+        """
+        class CSVImporter(Base):
+            date = Date(0)
+            narration = Column(1)
+            amount = Amount(2)
+            balance = Amount(3)
+            names = False
+            entry_order = EntryOrder.ReverseExecutionOrder
```
dnicolodi, line 396:
```suggestion
            order = Order.DESCENDING
```

beangulp/importers/csvbase.py
dnicolodi, line 177:
```suggestion
```suggestion
class Order(Enum):
    ASCENDING = 1
    """Entries are listed in chronological order."""
    DESCENDING = 2
    """Entries are listed in reverse chronological order."""
```
```

**Review by dnicolodi — CHANGES_REQUESTED** [review]
Thanks for working on this. This is just what I had in mind. I have mostly stylistic suggestions to make the code look more pythonic and one a little bit more substantial about not sorting lists if it is not needed. I haven't tested the suggestions, but I think any adjustment required should be straightforward.

**Comment by rbrunt** [comment]
Thanks for the review all very sensible suggestions. I've clearly been doing too much typescript recently 😅

**Comment by dnicolodi** [comment]
Can you please fix the lint error and squash the commits into one commit? The commit message could be something like:
```
importers.csvbase: Allow to specify source sort order and fix balances

Comparing the transactions date to determine the sort order does not
work when the source document contains transactions relative to just
one day. Make sure to use the right value for balance assertions
regardles of the source sort order.
```

dnicolodi:
Thanks, looks good! I just reworded the commit message title to make it fit into one line.

PR #148, opened 2025-01-10
Provide for passing through arguments to cli via Ingest

maread99:
This PR would allow users to pass through arguments to the click cli via the call to Ingest.

My particular case for requiring this is to be able to pass through the `standalone_mode` argument as False when testing. (This prevents the system from exiting after the execution instigated by the cli has finished. I might well be going about testing in the wrong way - I just simulate command line instructions).

I'd note that (at least some of) the arguments are passed through to at least `click.core.BaseCommand.main`. I haven't looked under-the-bonnet of the click decorators (on the `cli` function defined in `Ingest`) and don't know where else they might end up - I'm comfortable offering the PR given that I can't see how it could change any existing behaviour.

Anyway, one for you to consider.

Thank you as ever for all the work that continues to go into beancount and these supporting packages! It's very much appreciated.

### DIFF
--- a/beangulp/__init__.py
+++ b/beangulp/__init__.py
@@ -281,5 +281,5 @@ class Ingest:
 
         self.cli = cli
 
-    def __call__(self):
-        return self.cli()
+    def __call__(self, *args, **kwargs):
+        return self.cli(*args, **kwargs)


dnicolodi:
What's wrong with calling `Ingest.cli()` with the arguments you need instead than relying on the `__call__()` method? The `__call__()` method is just a convenience, and I would like to keep it as simple as possible to avoid possible confusion in users that are Python beginners.

maread99:
👍

Thanks for the reply Daniele.

For anyone who finds their way to this, for my specific case the `main` function looks like:
```python
def main():
    kwargs = {"standalone_mode": False} if TESTING else {}
    Ingest(my_importers, hooks=my_hooks).cli(**kwargs)
```

dnicolodi:
I may have misunderstood what you want to do, but there is a facility to test Click command line applications that makes the sort of thing you are doing in your last comment unnecessary, see https://click.palletsprojects.com/en/stable/testing/

You would write something like:
```python
import beangulp
import click.testing
ingest = beangulp.Ingest(importers)
runner = click.testing.CliRunner()
r = runner.invoke(ingest.cli, args)
assert r.exit_code == 0
assert r.output ...
```
See also the beangulp doctests, for example here https://github.com/beancount/beangulp/blob/2fab042a7336db92a9d1a13a78e7da29e761ff2b/beangulp/tests/extract.rst?plain=1#L22-L31

PR #151, opened 2025-01-20
Allow get_file to take _FileMemo input

jeremander:
Currently `beangulp.cache.get_file` will error if it is given a `_FileMemo` as input:

```
TypeError: expected str, bytes or os.PathLike object, not _FileMemo
```

This can be inconvenient in some situations; for example, the implementation of `CSVImporter.identify` is:

```python
def identify(self, filepath):
     return self.ident.identify(cache.get_file(filepath))
```

So this will error if `filepath` happens to already be a `_FileMemo`.

This makes the implementation of `get_file` slightly more flexible, to allow repeated applications of itself without causing an error.

### DIFF
--- a/beangulp/cache.py
+++ b/beangulp/cache.py
@@ -150,6 +150,8 @@ def get_file(filename):
       A FileMemo instance.
 
     """
+    if isinstance(filename, _FileMemo):
+        filename = filename.name
     assert path.isabs(filename), (
         "Path should be absolute in order to guarantee a single call.")
     return _CACHE[filename]
--- a/beangulp/cache_test.py
+++ b/beangulp/cache_test.py
@@ -30,6 +30,12 @@ class TestFileMemo(unittest.TestCase):
             self.assertEqual('abc', wrap.convert(converter))
             self.assertEqual(1, converter.call_count)
 
+            # Check that get_file can be called on a _FileMemo.
+            self.assertNotEqual(wrap, cache.get_file(tmpfile.name))
+            self.assertNotEqual(wrap, cache.get_file(wrap))
+            self.assertEqual(wrap.name, cache.get_file(tmpfile.name).name)
+            self.assertEqual(wrap.name, cache.get_file(wrap).name)
+
     def test_cache_head_and_contents(self):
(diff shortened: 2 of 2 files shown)


dnicolodi:
I don't think this is a good idea. Why do you want to call `beangulp.cache.get_file()` passing a `_FileMemo`? How do you end up calling `CSVImporter.identify()` with anything else other than a string? Can you paint the bigger picture? 

Furthermore, both `beangulp.cache` and `beangulp.importers.csv` are deprecated and I would prefer to eventually getting rid of them instead of piling more kludges on top.

dnicolodi:
The only thing you should be doing to make importers that work with `beancount.ingest` work with `beangulp` is to translate all imports from the `beancount.ingest` namespace to the `beangulp` namespace. Nothing more should be necessary. In particular, `get_file()` has most definitely not been changes, thus if that code (which I haven't looked at) works with beancount 2.3 it should also work with beangulp once you change the imports.

jeremander:
I think I was getting confused between the deprecated `beangulp.importers.csv.Importer` and the new `beangulp.importers.csv.CSVImporter`.  `Importer.identify` wants to take a `_FileMemo` as input, while `CSVImporter.identify` wants to take a string.  So you're right that replacing the old `beancount.ingest.importers.csv.Importer` with `beangulp.importers.csv.Importer` does the right thing (since both take a `_FileMemo`).

The issue is when I try to migrate from `Importer` to `CSVImporter`.  Until `beancount_import` updates itself to work with the new `beangulp.importer.Importer`, it will crash when it calls `identify`, because the input types are different.  So that's a bit unfortunate because it makes the new API backward incompatible.  Accepting a `_FileMemo` input in `get_file` would help avoid this particular crash.  Of course, no one said the new API has to be compatible with the old one; it just means that more things could potentially break if they don't update to the new one.

dnicolodi:
I don't follow your reasoning on why making the `_FileMemo` constructor accept a `_FileMemo` instance would help in moving an importer to the new importer interface: the new importer interface deals exclusively with strings to identify files, `_FileMemo` instances should not appear anywhere.

PR #152, opened 2025-01-23
feat: several improvements to csvbase.Importer

wzyboy:
I recently migrated from the legacy csv importer to beangulp but I find some functionality missing. I tried to add them back in a non-intrusive way.

This PR adds a few features to `csvbase.Importer`:

* add `invert_sign` argument to `Amount` column
* use `dateutil` to parse `Date` column if user does not specify a format (`dateutil` is already a dependency of Beancount)
* implement an `identify()` method based on the legacy csv importer's regexp matcher logic
* backport some of the legacy csv importer's column types as metadata keys if user chooses to define them as field names


dnicolodi:
> I recently migrated from the legacy csv importer to beangulp but I find some functionality missing. I tried to add them back in a non-intrusive way.

The `beangulp.imports.csvbase` importer is deliberately kept as simple as possible. It is however built in as modular as possible way to allow easy customization by sublassing the importer and other components. 

> * add `invert_sign` argument to `Amount` column

NAK. Please just subclass the `Amount` class:
```python
InvertedAmount(Amount):
    def parse(self, value):
        return -super().parse(value)
```
(from memory, the correct implementation may differ slightly).

> * use `dateutil` to parse `Date` column if user does not specify a format (`dateutil` is already a dependency of Beancount)

NAK. No guessing of date formats, it is error prone as the guessing depends on the value being parsed. Your bank statements do not change format: figure out the format once and encode it in your importer. IIRC, the `dateutil` dependency is there only because of the legacy CSV importer, which is on the chopping block.

> * implement an `identify()` method based on the legacy csv importer's regexp matcher logic

NAK. Implement the `identify()` method in your subclass: it takes less code than passing regular expression to a general method.

> * backport some of the legacy csv importer's column types as metadata keys if user chooses to define them as field names

NAK. Not generally useful enough. Implement the `finalize()` method.

wzyboy:
I have move logic for `identify()` and `metadata()` to subclasses.

However, I would like to contribue the changes to `Amount` and `Date` to upstream:

- I believe `invert_sign` is as useful and `subs`: Beancount uses negative amounts for Liabilities but most banks's credit card exports use positive amounts so the signs need inversion. I have 14 out of 23 importers that need this and it's also why it was implemented in the legacy CSV importer
- Some banks do not put year in the date. Using `datetime.datetime.strptime('12/31', '%m/%d')` returns 1900-12-31. Using `dateutil` avoids reinventing the wheels and solves a lot of headache if you have a lot of csv files to process (like me)

Would you review the reduced patch?

dnicolodi:
> * I believe `invert_sign` is as useful as `subs`: Beancount uses negative amounts for Liabilities but most banks's credit card exports use positive amounts so the signs need inversion. I have 14 out of 23 importers that need this and it's also why it was implemented in the legacy CSV importer

I may consider this, however the constructor parameter name needs to be named `negate`.

> * Some banks do not put year in the date. Using `datetime.datetime.strptime('12/31', '%m/%d')` returns 1900-12-31. Using `dateutil` avoids reinventing the wheels and solves a lot of headache if you have a lot of csv files to process (like me)

The problem with `dateutil.parser.parse()` is that it guesses the date format, often incorrectly. Most people in Europe would agree that "12/01/2025" and "13/01/2025" are two dates just one day apart. However
```python
>>> dateutil.parser.parse('12/01/2025') - dateutil.parser.parse('13/01/2025')
datetime.timedelta(days=322)
```
I all you want from `dateutil` is to use the current year for the missing value, you can do this:
```python
>>> datetime.datetime.strptime('23/01', '%d/%m').replace(year=datetime.date.today().year)
datetime.datetime(2025, 1, 23, 0, 0)
```
However, as you discovered yourself, this feature is not great if you are trying to import statements from a year that is not the current year. Your changes make it impossible to correctly import statements that are more that 180 days old and make the output of the importer not reproducible. Both these things are not acceptable in a general framework. Among other things this breaks the beangulp test functionality.

wzyboy:
I have renamed `invert_sign` to `negate`. Just a note that the legacy CSV importer, along with the adapter, still uses `invert_sign`.

Regarding of parsing the date, you made a point that trying to guess the year if the CSV does not have one is not reproducible so I removed that logic.

I used dateutil to parse the date here mainly because the original CSV importer used it. I hate the US-centric DMY date format as much as you do, but dateutil provides great convenience when I have dozens of CSV importers to deal with. There are so many ways to write the dates, e.g., `Jan`, `Jan.`, `January`, `JAN`, etc. `dateutil` reduces the headache to deal with all these combinations.

In its current state, `dateutil` is only used when the user does not explicitly specify a format. For ISO-8601 dates, `dateutil` never parses them wrong so it's totally backwards-compatible.

dnicolodi:
First, you seem to like very much how the legacy CSV importer works. Why don't you keep using it, instead than turning the new CSV import code into the legacy one? There are reasons why I developed the new CSV import code, one being that the legacy CSV importer is an unmaintainable mess because of the gazillion of options and special cases it implements. This is the main reason why I resist to add "just another little tweak" to the new CSV import base class.

> I used dateutil to parse the date here mainly because the original CSV importer used it. I hate the US-centric DMY date format as much as you do, but dateutil provides great convenience when I have dozens of CSV importers to deal with. There are so many ways to write the dates, e.g., `Jan`, `Jan.`, `January`, `JAN`, etc. `dateutil` reduces the headache to deal with all these combinations.

I explained why I don't like relying on guessing the date format. I'm not going to accept a patch that makes the main column date parser to guess the date format. However, the design of the CSV importer base class is composable. If you like the guessing, you can implement your `GuessDateFormatAndBreakUnexpectedly` column format class and use it in your importers.

> In its current state, `dateutil` is only used when the user does not explicitly specify a format. For ISO-8601 dates, `dateutil` never parses them wrong so it's totally backwards-compatible.

It is not backward compatible: now for malformed dates an error is raised, with the change you get a best guess of what the date is. The two things are most definitely not the same.

I think the change to the `Amount` column parser is ok. However, there are still unrelated changes in the PR and you should squash all commits into one or a few implementing one logical change each, with a suitable commit message.

wzyboy:
I have removed `dateutil` and squashed commits as you requested.

beangulp/importers/csvbase.py
```diff
@@ -121,6 +122,8 @@ def __init__(self, name, frmt='%Y-%m-%d'):
         self.frmt = frmt
 
     def parse(self, value):
+        if not value:
+            return None
```
dnicolodi, line 126: Why is returning `None` acceptable here? If the input is not a valid date an error should be raised.

beangulp/importers/csvbase.py
```diff
@@ -137,17 +140,24 @@ class Amount(Column):
       subs: Dictionary mapping regular expression patterns to
         replacement strings. Substitutions are performed with
         re.sub() in the order they are specified.
+      negate: If true, flip the sign of the amount.
```
dnicolodi, line 143:
```suggestion
      negate: If True, negate the amount.
```

beangulp/importers/csvbase.py
```diff
@@ -137,17 +140,24 @@ class Amount(Column):
...
 
-    def __init__(self, name, subs=None):
+    def __init__(self, name, subs=None, negate=False):
         super().__init__(name)
         self.subs = subs if subs is not None else {}
+        self.negate = negate
 
     def parse(self, value):
+        if value == "":
+            return None
```
dnicolodi, line 154: Why?

beangulp/importers/csvbase.py
```diff
@@ -291,35 +301,39 @@ def extract(self, filepath, existing):
...
+                                    row.date, flag, payee, row.narration, tags, links, [
+                                        data.Posting(account, units, None, None, None, None),
+                                    ])
+
+                # Apply user processing to the transaction.
+                txn = self.finalize(txn, row)
+                if txn is None:
+                    continue
+            except Exception:
+                print(f'Error processing file {filepath}, line {lineno}: {row!r}', file=sys.stderr)
```
dnicolodi, line 335: Printing to standard error is not the way to do this.

dnicolodi (changes requested):

beangulp/importers/csvbase.py
dnicolodi: Please note that the `Column` base class already has a mechanism for returning a default value when the value is empty. However, it is not exposed by the `Amount` subclass. It probably should. And the mechanism should be extended to make `None` a valid default. However, making returning `None` on empty values the default behavior of the `Amount` column description does not seem to be a good idea to me.

beangulp/importers/csvbase.py
wzyboy: A CSV file can have multiple Date columns, e.g. posted date and transaction date. The posted date is used for `entry.date` and the transaction can be added to `entry.meta` in `finalize()`. But such a Date column might be empty for some transactions (e.g. `PAYMENT RECEIVED`). If an error is raised here, user is unable to process such rows in `finalize()`.

beangulp/importers/csvbase.py
wzyboy: A CSV file can have multiple Amount columns, e.g. one for "DEBIT" and another for "CREDIT". Only one of the two has value and the other is an empty string. `Decimal('')` raises exception so user cannot further process the row in `finalize()`. So I'm allowing empty string as Amount here.

Do you think `Decimal('0')` is a better idea than `None`?

beangulp/importers/csvbase.py
wzyboy: Can you suggest a better way? My intention was to make it easier to debug which row causes parsing error when dealing with all kinds of not-so-standard CSV files.

Currently it's just a generic "importer error" without reavling the culprit row. It would be better to include this information in the traceback, but I don't know how.

beangulp/importers/csvbase.py
dnicolodi: Sure, that is a valid use case, but it should not be what the column adapter does by default. There must be a `default` parameter that can be passed to the constructor to enable returning a default when the value is the empty string. See how this is done for the generic `Column` class. That mechanism should be exposed here and extended to allow to use `None` as a default value. I'll implement this ASAP.

beangulp/importers/csvbase.py
dnicolodi: The default value should be specified by the user.

However, for statements that have a "debit" and a "credit" columns, you most likely want something different, see the code proposed here https://github.com/beancount/beangulp/pull/126#issuecomment-2211999829 I don't remember testing it, but it should work with little adjustments.

beangulp/importers/csvbase.py
dnicolodi: The proper way to do this is to chain exceptions. I'll implement this.

dnicolodi:
I've implemented the default value idea and I cherry picked the amount negation patch from this PR and committed that separately. I rebased this PR on top. I'll look at the other patches later.

beangulp/importers/csvbase.py
wzyboy: Thanks for implementing the default value. It works great!

beangulp/importers/csvbase.py
wzyboy: Thanks. The `WithdrawlOrDeposit` idea is awesome.

dnicolodi: If it is useful, we can add it to beangulp. The only difficult thing is to find a suitable name. I don't like `WithdrawalOrDeposit`, `DebitOrCredit` is a bit better, but not much. Does anyone have a better idea?

wzyboy: `DebitOrCredit` sounds good to me. If you want a new name...

... since we have `Columns`, how about `Amounts`? It takes multiple columns, and finds the non-empty one.

But as `Columns` concats strings together, `Amounts` might make people think that it adds up decimals together.

dnicolodi: Added in #155.

dnicolodi (approved):
Thanks for your contribution @wzyboy. I adjusted a few things and force pushed to your branch. Please have a look and let me know what you think.

wzyboy (commented):
Thank you @dnicolodi ! LGTM!

PR #159, opened 2025-03-06
Don't merge deduped entries into existing_entries.

hoostus:
Merging pollutes the existing_entries for hooks that get run later. Instead we track them separately but consider the union(existing_entries, new_entries) when running the deduplication logic, since we also want to remove duplicates from other importers in this same.

### DIFF
--- a/beangulp/__init__.py
+++ b/beangulp/__init__.py
@@ -96,9 +96,12 @@ def _extract(ctx, src, output, existing, reverse, failfast, quiet):
     extract.sort_extracted_entries(extracted)
 
     # Deduplicate.
+    deduplicated_new_entries = []
     for filename, entries, account, importer in extracted:
-        importer.deduplicate(entries, existing_entries)
-        existing_entries.extend(entries)
+        # We want to deduplicate not just against existing_entries
+        # but also new entries from other importers in this run
+        importer.deduplicate(entries, existing_entries + deduplicated_new_entries)
+        deduplicated_new_entries.extend(entries)
 
     # Invoke hooks.
     for func in ctx.hooks:


hunner:
Agreed -- I came to report this issue as well because of an effect I saw with `smart_importer`.

When new entries are merged with existing ones, the combined set is passed to `smart_importer` as training data. This includes single-leg entries, which shouldn’t be considered valid training examples. As a result, `smart_importer` is producing single-leg entries in my journal, since it learns from these incomplete examples.

dnicolodi:
> I guess if same input files are kept but the order of `importers` is changed then that affects the `existing_entries` that each successive `importer.deduplicate` in the deduplication for-loop sees.

I guess you didn't read the large comment in the code that I linked that describes the sorting key used.

hoostus:
> the prior behavior seems reasonable to me

I dunno, it means that hooks are being passed an invalid beancount file as "existing entries", which is at least slightly surprising. I don't think it is unreasonable that hooks would expect the existing entries to be something that would pass bean-check. If hooks are expected to deal with malformed beancount data then we should probably clarify what variants of well-formed beancount syntax they might be expected to deal with in the "existing entries" so they can write appropriately defensive code.

I'm not exactly averse to any of that -- importing stuff is kind of a weird liminal state instead of fully baked -- but, as someone who wrote a hook, I was caught out by beangulp breaking the (admittedly somewhat implicit) contract that "existing entries" were well-formed from a bean-check kind of point of view. I expected it in the imported data but not the existing data.

dnicolodi:
Thanks for spotting this issue and sorry for not replying on the PR earlier: I had to find a bit of time to check what the code does against what it is supposed to be doing and what it is sensible to do. I agree that the current behavior is a bug. Newly extracted entries should not be appended to the existing entries before these are passed to the hook functions.

I have only one issue with the proposed fix: it repeatedly concatenates possibly very long lists. I need to think a bit more about it, but I think that the deduplication process should be additive, namely `deduplicate(A, B + C)` should give the exact same result of `deduplicate(A, B); deduplicate(A, C)` where the first argument is the list of new entries and the second argument is the list of existing entries. Thus it may be more efficient to run consecutive `dedup()` calls than concatenating and sorting lists.

Issue #134, opened 2024-11-11
pytest compatibility


andylou2:
Hi, 

In beancount v2 there is a handy utility to incorporate Importer unit tests with pytest: https://github.com/beancount/beancount/blob/v2/beancount/ingest/regression_pytest.py

I couldn't find the equivalent in beangulp. In the examples directory, I saw the tests are driven by a bash script: https://github.com/beancount/beangulp/blob/master/examples/importers/runtests.sh

Are you guys open to adding the pytest compatibility to beangulp, or accepting PRs? Something along the lines of this:
```
import os
from typing import List, Optional, Union

import beangulp
import beangulp.testing
import click
import pytest
from beangulp import utils
from beangulp.importer import Importer, ImporterProtocol


def run_test_programmatically(
    importer: Union[Importer, ImporterProtocol],
    documents: List[str],
    expected: Optional[str] = None,
    verbose: int = 0,
    quiet: int = 0,
    failfast: bool = False,
    generate: bool = False,
) -> None:
    """
    Programmatically run the test command that would normally be invoked via CLI.

    Args:
        importer: The importer instance to test
        documents: List of document paths to test
        expected: A directory for expected output files
        verbose: Verbosity level (default: 0)
        quiet: Quiet level (default: 0)
        failfast: Whether to stop at first failure (default: False)
        generate: If True, then generates test data
    """
    ctx_obj = beangulp.Ingest([importer])
    command = beangulp.testing._generate if generate else beangulp.testing._test

    with click.Context(command, obj=ctx_obj) as ctx:
        if generate:
            command.callback(documents=documents, expected=expected, verbose=verbose, quiet=quiet, force=True)

            # Verify that the expected files were generated
            for doc in utils.walk(documents):
                if doc.endswith(".beancount"):
                    continue

                expected_filename = f"{doc}.beancount"
                if not os.path.exists(expected_filename):
                    raise FileNotFoundError(f"Did not generate expected file: {expected_filename}")
        else:
            command.callback(documents=documents, expected=expected, verbose=verbose, quiet=quiet, failfast=failfast)
```

Thanks,
Andy

blais:
SGTM

PR #163, opened 2025-03-29
Taking a stab at bulk testing importers.

alexrkaufman:
I used the code from `testing.py` as inspiration but tried to match the approach used in `_extract`

This adds a commands to the cli for `test` and `generate` that can use the same importer configurations as you would use for typical use.

I dont know that this is necessarily a good approach and would love feedback on how I could improve this or alternate paths for making testing custom beangulp importers easier.

Seems related to https://github.com/beancount/beangulp/pull/118 and https://github.com/beancount/beangulp/issues/134.


alexrkaufman:
the force push was just to fix changes I accidentally included that were accidentally imposed by my formatter which should make it clearer what I actually changed. 

Hope this helps!

dnicolodi:
Can you explain which problem this solves? At a quick glance it looks a copy of the commands implemented in the `beangulp.testing` module. Why is it needed? If you want to test more than one importer at a time, just iterate over the importers in a script or a makefile.

alexrkaufman:
Hi! Thank you for getting back to me and taking a look at this. 

The problem this is meant to solve is to make testing more straightforward for users and enable testing with minimal (if any) modification to their `import.py` files. 

You are right that this is very similar to `testing.py`. I modified that approach which enforces only one importer in the context to allow multiple importer. I made the `test` and `generate` commands part of the same cli as the identify, extract, and archive commands so the use becomes: 

```python
./import.py generate test_documents/ 
```

Would it be better to just modify the code that is already implemented in `testing.py` and then import it into `__init__.py` so that it exposes the `test` and `generate` commands? 

It is possible this is overfit for my use case but below are some benefits I felt this approach had over a script like `runtests.sh`

* It doesnt require writing an additional script. 
* This also allows me to use my archive folder as the test documents folder.
* `runtests.sh` loops over each importer and checks every file in the given folder. What I wrote uses the same identification method as `identify` and `extract`.
* This approach does not requires duplicating configuration already implemented in `import.py`

I have made previous unfruitful attempts to get testing set up and spent the better part of today going through the mailing list, examples here, old beancount.ingest documentation, and reds importers trying to figure this out. It is very likely that there are concepts or core ideas I am not aware of. Those other issues I linked to seem like they may also solve a similar problem and probably are suggesting something better but I don’t understand them so I tried to do what I was able to figure out. If there is a better approach to pursue with pytest or something please let me know and I would be happy to pursue that!

alexrkaufman:
Hello, just following up here. I hope the explanation of the situation I was in makes sense. If this is not a good approach please close this pull request and let me know if there is a better thread to pull like getting #134 to a state where it could be included in the main branch. 

Greatly appreciate all the work on the project so far and would love to contribute.

PR #165, opened 2025-04-30
Support CSV files where the amount value embeds currency symbol, for example, "$123,456.78"

rajive:


pwalkr:
I've worked around this using the regex map option for amount:
```
    amount = csvbase.Amount('Amount (total)', {
        r'\+ \$': '',
        r'- \$': '-',
        r'^$': '0',
        })
```
Given `utils.parse.amount` only supports USD `$`, what about also having this line just remove any non-`[-.0-9]`?
https://github.com/beancount/beangulp/blob/cca19d5880bd71e44c96801e79adf2370eeee1b9/beangulp/utils.py#L125

dnicolodi:
> I've worked around this using the regex map

Why do you call this a work around? It is the intended use of the class.

PR #180, opened 2025-06-26
Allow ingest() to exit normally rather than calling sys.exit(0).

boogiewookie:
Allow passing standalone_mode=True to Ingest.__call__() so it exits normally rather than calling sys.exit(0). That
allows further processing afterward. In my use case, after extracting  the transactions from ofx/csv, I want to extract balances from pdfs for comparison, then split the output lines into multiple journal files.

I also tweaked the Makefile as it was complaining .../uv/builds-v0/.tmp88uxMu/bin/python: No module named pytest

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,9 +3,11 @@
 # Test in an isolated Python environment.
 # uv python pin <version> to set the version.
 test:
-	(env -u PYTHONPATH  uv run --isolated --python 3.13  python -m pytest)
+	(env -u PYTHONPATH uv run --isolated --python 3.13 \
+		--with petl --with pytest Python -m pytest)
 
 test-all:
 	for V in 3.7 3.8 3.9 3.10 3.11 3.12 3.13 ; do \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -3,9 +3,11 @@
 # Test in an isolated Python environment.
 # uv python pin <version> to set the version.
 test:
-	(env -u PYTHONPATH  uv run --isolated --python 3.13  python -m pytest)
+	(env -u PYTHONPATH uv run --isolated --python 3.13 \
+		sh -c "uv pip install petl pytest && python -m pytest")
```
blais, line 7: Don't do this; instead use `uv run --with petl` or somesuch.

blais:
LGTM. Please make the one change I'm requesting (in both places).

Makefile
boogiewookie: Thanks! I wasn't previously aware of --with. That's so much cleaner.

PR #189, opened 2025-09-12
add py.typed marker file

trim21:
this will at least help pylance(vscode python extension) to not warn about missing stubs of beangulp


blais:
Can you fix th elin error?

PR #191, opened 2025-10-01
Modify identify function to return Optional[Importer]

Ulthran:
Updated the identify function signature to return an Optional Importer.


beangulp/extract.py
```diff
@@ -36,7 +36,7 @@
 
 def extract_from_file(
     importer: "Importer", filename: str, existing_entries: "data.Directives"
-) -> data.Entries:
+) -> data.Directives:
```
dnicolodi: `data.Entries` and `data.Directives` are the same thing

Issue #197, opened 2026-01-01
Make the CSV Base Importer's date function robust against empty lines


johannesjh:
Please make the `beangulp.importers.csvbase.Importer.extract` method robust against empty lines in the CSV file. E.g., by simply adding `if row` to the expression, resulting in `return max(row.date for row in self.read(filepath) if row)`.

**Problem**: The CSV Base Importer's date function currently is not robust against empty lines in the CSV files. In other words, empty lines in the CSV lead to an `IndexError` exception as follows:

```txt
HTTP 500 - Error calling method on importer:

Traceback (most recent call last):
  File ".venv/lib/python3.13/site-packages/fava/core/ingest.py", line 183, in wrapper
    return func(*args, **kwds)
  File ".venv/lib/python3.13/site-packages/fava/core/ingest.py", line 237, in file_import_info
    date = importer.date(str_path)
  File ".venv/lib/python3.13/site-packages/beangulp/importers/csvbase.py", line 264, in date
    return max(row.date for row in self.read(filepath))
  File ".venv/lib/python3.13/site-packages/beangulp/importers/csvbase.py", line 264, in <genexpr>
    return max(row.date for row in self.read(filepath))
               ^^^^^^^^
  File ".venv/lib/python3.13/site-packages/beangulp/importers/csvbase.py", line 72, in func
    value = tuple(obj[i] for i in idxs)
  File ".venv/lib/python3.13/site-packages/beangulp/importers/csvbase.py", line 72, in <genexpr>
    value = tuple(obj[i] for i in idxs)
                  ~~~^^^
IndexError: tuple index out of range
```

**Analysis**: The current implementation in beangulp looks like this, see ` beangulp.importers.csvbase.Importer.date`. The code expects each row to have a "date" attribute, but this expectation is not met by rows resulting from empty lines in the .csv file.

```python3
    def date(self, filepath):
        """Implement beangulp.Importer::date()"""
        return max(row.date for row in self.read(filepath))
```

On a side note, other parts beangulp's CSV Base Importer are already robust against empty lines. Specifically, `beangulp.importers.csvbase.Importer.extract` includes the following lines to handle empty lines:

```python3
            # Skip empty lines.
            if not row:
                continue
```

**Suggested solution**: Make `beangulp.importers.csvbase.Importer.date` robust against empty lines, in a similar way to how the extract method already is robust against empty lines. This can by achieved by adding just the two words `if row` to the date function, as follows:

```python3
    def date(self, filepath):
[...]

PR #199, opened 2026-01-04
Allow to customize raw CSV file reading

### LINKED ISSUE #196: Preprocessing a file for csvbase.Importer requires acrobatics with temporary files
As CSV files downloaded from banks are notoriously messy, one might need to pre-process them beyond overwriting the `header` and `footer` members of 'csvbase.CSVReader`. For example. there might be a variable number of header lines. The common solution is to read in the file as a string, do the preprocessing (e.g. skip all lines up to a keyword, e.g. the first column header), then provide the updated string to the usual processing. 

But `csvbase.CSVReader.read()` expects a file name as argument, not a file-like object or a content string. Therefore, one must do something like this

```python
class MyMessyBank:
  
  def read(self, filepath):
    with open(filepath, 'r') as f:
      content = f.read()
    content = my_preprocessing(content)
    with tempfile.NamedTemporaryFile as t:
      t.writelines(content)
      for row in super().read(t.name):
        yield row
```

it would be easier if one could just use `StringIO()` or similar to provide modified data to `CSVReader.read()`.

There are multiple options that come to my mind

* Allow `read(str, ...)` as well as `read(file, ...)`
* Introduce `read()` usage like `read(None, ... , content: <file>)`
* Add a `open(filename) -> file` function to `importers.Importer` that can be overwritten by subclasses

mlell:
While CSVReader has numerous options to skip header or footer lines, or skip lines based on prefixes, it is cumbersome to read lines with varying lines before or after the actual statements. Users would need to override CSVReader.read(). Simply extending using `super().read(...)` is possible but requires creating a temporary file because read() needs a physical file path. Simply adding an `open(str) -> file-like` method to CSVReader allows to keep the useful functionality of `read()` while still being able to preprocess the file, for example, skipping a variable number of lines based on a pattern. 

This would fix #196.

### DIFF
--- a/beangulp/importers/csvbase.py
+++ b/beangulp/importers/csvbase.py
@@ -276,6 +276,25 @@ class CSVReader(metaclass=CSVMeta):
             # warnings.warn('skiplines is deprecated, use header instead', DeprecationWarning)
             self.header = self.skiplines
 
+    def open(self, filepath, encoding):
+        """Open filepath and return an iterable yielding lines of CSV-formatted text.
+        
+        This method can be overridden in subclasses to customize file reading, for 
+        example to pre-proceess text lines before import. To simply skip a fixed
+        number of lines at the beginning or end, consider setting the ``header`` 
+        or ``footer`` class variables instead.
+
+        Args:
+          filepath: Filesystem path to the input file.
+          encoding: File encoding as provided to the ``open()`` built-in
+            function.
+
+        Returns:
+          An iterable providing lines of CSV-formatted text.
+        """
+        with open(filepath, encoding=encoding) as fd:
+            yield from fd
+
     def read(self, filepath):
         """Read CSV file according to class defined columns specification.
 
@@ -292,34 +311,35 @@ class CSVReader(metaclass=CSVMeta):
 
         """
 
-        with open(filepath, encoding=self.encoding) as fd:
-            # Skip header and footer lines.
-            lines = _chomp(fd, self.header, self.footer)
+        lines = self.open(filepath, self.encoding)
 
-            # Filter out comment lines.
-            if self.comments:
-                lines = filter(lambda x: not x.startswith(self.comments), lines)
+        # Skip header and footer lines.
+        lines = _chomp(lines, self.header, self.footer)
 
-            reader = csv.reader(lines, dialect=self.dialect)
+        # Filter out comment lines.
+        if self.comments:
+            lines = filter(lambda x: not x.startswith(self.comments), lines)
 
-            # Map column names to column indices.
-            names = None
-            if self.names:
-                headers = next(reader, None)
-                if headers is None:
-                    raise IndexError("The input file does not contain an header line")
-                names = {name.strip(): index for index, name in enumerate(headers)}
+        reader = csv.reader(lines, dialect=self.dialect)
 
-            # Construct a class with attribute accessors for the
-            # configured columns that works similarly to a namedtuple.
-            attrs = {}
-            for name, column in self.columns.items():
-                attrs[name] = property(column.getter(names))
-            row = type("Row", (tuple,), attrs)
+        # Map column names to column indices.
+        names = None
+        if self.names:
+            headers = next(reader, None)
+            if headers is None:
+                raise IndexError("The input file does not contain an header line")
+            names = {name.strip(): index for index, name in enumerate(headers)}
 
-            # Return data rows.
-            for x in reader:
-                yield row(x)
+        # Construct a class with attribute accessors for the
+        # configured columns that works similarly to a namedtuple.
+        attrs = {}
+        for name, column in self.columns.items():
+            attrs[name] = property(column.getter(names))
+        row = type("Row", (tuple,), attrs)
+
+        # Return data rows.
+        for x in reader:
+            yield row(x)
 
 
 class Importer(beangulp.Importer, CSVReader):
--- a/beangulp/importers/csvbase_test.py
+++ b/beangulp/importers/csvbase_test.py
@@ -2,6 +2,7 @@ import datetime
 import decimal
 import re
 import unittest
+from itertools import dropwhile
 
 from beancount.core import data
 from beancount.parser import cmptest
@@ -450,6 +451,31 @@ class TestCSVReader(unittest.TestCase):
         rows = list(reader.read(filename))
         self.assertEqual(len(rows), 1)
         self.assertEqual(rows[0][0], "a")
+
+    @docfile
+    def test_custom_open(self, filename):
+        """\
+        Skip this line
+        Skip this too
+        First, Second
+        a, b
+        c, d
+        """
+
+        class Reader(CSVReader):
+            first = Column("First")
+            second = Column("Second")
+
+            def open(self, filepath):
+                """Skip lines until we find the column headers."""
+                lines = super().open(filepath)
+                return dropwhile(lambda line: "First" not in line, lines)
+
+        reader = Reader()
+        rows = list(reader.read(filename))
+        self.assertEqual(len(rows), 2)
+        self.assertEqual(rows[0].first, "a")
+        self.assertEqual(rows[1].second, "d")
 
 
 class Base(Importer):


johannesjh:
I second the proposition. It would also help with #197 

Instead of strictly requiring a file-like object, the importer could allow more general `Iterable[String]` type of objects. Users could still pass file-like objects, but also generator expressions. Generator expressions could be a convenient and idiomatic way of manipulating / filtering CSV lines.

beangulp/importers/csvbase.py
```diff
@@ -284,6 +302,10 @@ def read(self, filepath):
         accessors to access the data fields as defined by the class
         columns specification.
 
+        This method uses the class members header, footer, names, dialect,
+        comments, and order. Overriding this method causes those members
+        to be ignored unless the overriding method explicitly uses them.
+
```
dnicolodi: This change does not belong to this commit or this PR. I also find it unnecessary.

beangulp/importers/csvbase.py
```diff
@@ -276,6 +276,24 @@ def __init__(self):
             # warnings.warn('skiplines is deprecated, use header instead', DeprecationWarning)
             self.header = self.skiplines
 
+    def open(self, filepath):
+        """Open the CSV file for reading.
+
+        This method can be overridden in subclasses to customize raw file reading,
+        for example to skip lines before processing or to handle special file formats.
```
dnicolodi: What does "special file formats" mean here? This class supports a declarative way to define an importer for CSV files. Why would you want to use it for something else?

dnicolodi (changes requested):

beangulp/importers/csvbase.py
```diff
@@ -276,6 +276,24 @@ def __init__(self):
...
+        This method uses the class member 'encoding'. Overriding this method causes
+        that member to be ignored unless the overriding method explicitly uses it.
+
+        Args:
+          filepath: Filesystem path to the input file.
+
+        Returns:
+          A file object opened for reading.
+
+        """
```
dnicolodi, line 294: I agree with @johannesjh: this interface is not optimal. Having the `open()` method return an `Iterable[str]` would be better. A
```python
def iterlines(self, fd):
    yield from fd
```
and use it where appropriate, would be another, maybe even better, possibility.

beangulp/importers/csvbase_test.py
```diff
@@ -451,6 +452,38 @@ class Reader(CSVReader):
...
+    def test_custom_open(self, filename):
+        """\
+        Skip this line
+        Skip this too
+        First, Second
+        a, b
+        c, d
+        """
+
+        class CustomReader(CSVReader):
```
dnicolodi:
```suggestion
        class Reader(CSVReader):
```
Please keep the classes in the tests named uniformly.

beangulp/importers/csvbase_test.py
```diff
@@ -451,6 +452,38 @@ class Reader(CSVReader):
...
+                """Skip lines until we find the column headers."""
+                fd = super().open(filepath)
+                # Read lines until we find one containing "First"
+                for line in fd:
+                    if "First" in line:
+                        # Create a new file-like object with the header line and remaining content
+                        remaining = fd.read()
+                        fd.close()
+                        return io.StringIO(line + remaining)
+                return fd
```
dnicolodi: With the `iterlines()` idea exposed above, this would become simply:
```python
from itertools import dropwhile

class Reader(CSVReader):
    first = Column("First")
    second = Column("Second")

    def iterlines(self, fd):
        return dropwhile(lambda line: "First" not in line, fd)
```
which seems much better to me.

dnicolodi (changes requested):

beangulp/importers/csvbase.py
mlell: The rationale was that there would be two ways to customize importer behaviour which partially overlap. Overriding "open()" would render the member "encoding" non-functional, for example. Will remove it.

beangulp/importers/csvbase.py
mlell: I see that this is unclearly phrased. I wanted to refer to the fact that CSV has a large number of strange variants.

mlell: Also, there are grey areas I think, like compressed CSV or XLSX which are not strictly CSV but are conceptually similar and can easily be converted

beangulp/importers/csvbase_test.py
mlell: Reader class is [now](https://github.com/beancount/beangulp/blob/a5561e397cd6cc0c6335ab75ff256441165d9f51/beangulp/importers/csvbase_test.py#L465-L472):

```py
        class Reader(CSVReader):
            first = Column("First")
            second = Column("Second")


            def open(self, filepath):
                """Skip lines until we find the column headers."""
                lines = super().open(filepath)
                return dropwhile(lambda line: "First" not in line, lines)
```

beangulp/importers/csvbase_test.py
mlell: done

beangulp/importers/csvbase.py
mlell: Commit [a5561e3](https://github.com/beancount/beangulp/pull/199/commits/a5561e397cd6cc0c6335ab75ff256441165d9f51) has `open() -> Iterable[str]`

beangulp/importers/csvbase.py
mlell: done in [a5561e3](https://github.com/beancount/beangulp/pull/199/commits/a5561e397cd6cc0c6335ab75ff256441165d9f51)

mlell:
* Changed the return type of `open()` to be an iterable (e.g., a generator) to allow for e.g. generator expressions
* Removed the change to the docstring of `read()`
* Updated the unit test to use the `dropwhile` approach

While it is true that this is a reader for CSV data only, I felt that there might be some formats, like compressed data or XLSX, where it might be useful to first open in binary mode, preprocess and then hand over to `read()`. This would not be possible with the `iterlines()`. Approach. However, you might have experience with more different bank statement formats that I have, if you think noone is going to need this, the `iterlines()` approach would be simpler.

beangulp/importers/csvbase.py
```diff
@@ -276,6 +276,26 @@ def __init__(self):
...
+    def open(self, filepath):
+        """Open the CSV file for reading.
+
+        This method can be overridden in subclasses to customize raw file reading,
+        for example to pre-proceess text lines before import. Note that to skip
+        a fixed number of lines at the file beginning or end, setting the class
+        members "header" or "footer" is the easier approach.
+
+        This method uses the class member 'encoding'. Overriding this method causes
+        that member to be ignored unless the overriding method explicitly uses it.
```
dnicolodi:
```suggestion
        """Open filepath and return an iterable yielding lines of CSV-formatted text.
        
        This method can be overridden in subclasses to customize file reading, for 
        example to pre-proceess text lines before import. To simply skip a fixed
        number of lines at the beginning or end, consider setting the ``header`` 
        or ``footer`` class variables instead.
```

beangulp/importers/csvbase.py
```diff
@@ -276,6 +276,26 @@ def __init__(self):
             # warnings.warn('skiplines is deprecated, use header instead', DeprecationWarning)
             self.header = self.skiplines
 
+    def open(self, filepath):
```
dnicolodi:
```suggestion
    def open(self, filepath, encoding):
```

beangulp/importers/csvbase.py
```diff
@@ -276,6 +276,26 @@ def __init__(self):
...
+        This method uses the class member 'encoding'. Overriding this method causes
+        that member to be ignored unless the overriding method explicitly uses it.
+
+        Args:
+          filepath: Filesystem path to the input file.
+
+        Returns:
+          An iterable providing lines of CSV-formatted text.
+        """
+        with open(filepath, encoding = self.encoding) as fd:
```
dnicolodi:
```suggestion
        with open(filepath, encoding=encoding) as fd:
```

beangulp/importers/csvbase.py
```diff
@@ -292,34 +312,35 @@ def read(self, filepath):
...
-            # configured columns that works similarly to a namedtuple.
-            attrs = {}
-            for name, column in self.columns.items():
-                attrs[name] = property(column.getter(names))
-            row = type("Row", (tuple,), attrs)
-
-            # Return data rows.
-            for x in reader:
-                yield row(x)
+        lines = self.open(filepath)
```
dnicolodi:
```suggestion
        lines = self.open(filepath, self.encoding)
```

dnicolodi (changes requested):
Better but not needs some tweaks.

mlell:
I have added your suggestions and rebased the branch to the latest version! What do you think @dnicolodi ?

floli:
Kindly pinging on that PR. Is there something I can do to support?

PR #200, opened 2026-01-06
Fix crash in import.py when there are multiple slashes in entry

clouserw:
`examples/import.py` currently crashes when the entry has more than one slash.  For example, I ran into this:

> COMCAST / XFINITY        800-266 / COMCAST / XFINITY 800-266-2278 WA

This patch splits on the last slash which is a compromise but better than splitting on the first, I think.


dnicolodi:
The example is just an example that should handle only the example data. How do you run into a issue with it?

clouserw:
People use examples as templates.  I copied it to use on other data and ran into the problem.

